Issue #144, opened 2026-05-05
Try --via-ir compilation to reduce StoxReceiptVault size


thedavidmeister:
Sub-task of #90.

Try compiling with `--via-ir` (route 1 instead of route 0) and measure the size delta on `StoxReceiptVault`. Currently 26,134 runtime / 1,558 over EIP-170 with the legacy pipeline.

## Steps

1. Add `via_ir = true` under `[profile.default]` in `foundry.toml`.
2. Re-run the size check (`forge build --sizes`). Record before/after for every contract, not just the vault — IR can shift sizes elsewhere too.
3. Run the full suite. IR sometimes surfaces stack-too-deep errors that the legacy pipeline papered over with optimizer help.
4. If size + tests both improve / stay green: regenerate every pointer (`script/BuildPointers.sol`), update `LibProdDeploy*` constants, re-run fork-codehash tests.
5. CHANGELOG entry under V3 (or V4 if pointer values shift across the deploy set).

## Expected outcome

IR typically saves 5–15% bytecode on contracts with deep inheritance / heavy inlining. Vault has both, so likely toward the high end. If the saving lands at >1,558 bytes the contract becomes deployable; if not, fall back to the other levers in #90 (move logic to facet, lower `optimizer_runs`, etc.).

## Risks

- IR codegen is a different pipeline than the legacy one. Even with green tests, every code path has new bytecode characteristics. Worth a slither pass on the IR-compiled output.
- Codehash cascade: every pointer file changes. The corporate-actions PR stack already exercised this cascade; same drill.
- IR is theoretically deterministic but has had non-determinism bugs historically — pin compiler version (`solc = "0.8.25"` already exact) and double-check pointer regen converges.

## Discard signal

If size delta is <500 bytes, IR alone won't close the gap and it's not worth the per-call gas cost (IR sometimes costs slightly more runtime gas than legacy at high `runs`). Move on to facet extraction or `optimizer_runs` reduction without merging the IR switch.

thedavidmeister:
Tried, won't merge. IR makes the vault **larger**, not smaller.

Measured with `via_ir = true` added to `foundry.toml`:

| Contract | Legacy | IR | Δ |
|---|---|---|---|
| StoxReceiptVault | 26,134 | **27,986** | **+1,852 (worse)** |
| TestStoxReceiptVault | 27,699 | 28,354 | +655 (worse) |
[...]

Issue #90, opened 2026-04-21
StoxReceiptVault runtime exceeds EIP-170 by 1,174 bytes


thedavidmeister:
`forge build --sizes` reports StoxReceiptVault runtime at 25,750 bytes, 1,174 over the EIP-170 limit of 24,576. The contract cannot be deployed via CREATE/CREATE2 to any chain enforcing EIP-170 in its current state.

```
| Contract         | Runtime | Init   | Margin  |
|------------------|---------|--------|---------|
| StoxReceiptVault | 25,750  | 26,006 | -1,174  |
```

CI does not catch this because the test command does not check bytecode size. Only `forge build --sizes` surfaces it.

The over-limit state was introduced as `StoxReceiptVault` inlined `LibRebase.migratedBalance`, `LibTotalSupply.{fold,onMint,onBurn,onAccountMigrated,effectiveTotalSupply}`, `LibERC20Storage.{underlyingBalance,setUnderlyingBalance,underlyingTotalSupply}`, `LibCorporateAction.getStorage`, and the `AccountMigrated` emit into its own runtime.

Blocks deployment. Needs to be resolved before PR #24 can reach production.

thedavidmeister:
Status update (2026-05-05): worse than originally filed.

`forge build --sizes` now reports:

| Contract | Runtime | Init | Margin (EIP-170) |
|---|---|---|---|
| StoxReceiptVault | **26,134** | 26,390 | **-1,558** |
| TestStoxReceiptVault | 27,699 | 27,955 | -3,123 |

Was `-1,174` over when this issue was filed; the corporate-actions stack landings since then added more inlined logic, pushing the margin to `-1,558`. The contract cannot be deployed via CREATE/CREATE2 on any chain that enforces EIP-170 (every L1 + every major L2 — Ethereum mainnet, Base, Arbitrum, Optimism, Polygon all 24,576).

### Why CI doesn't catch it

`rainix-sol-test` and `rainix-sol-static` run the test suite + slither + fmt-check, none of which check runtime bytecode size. Only `forge build --sizes` surfaces it; it isn't in any CI script today.

### Why fork-codehash tests pass despite the overflow

Worth investigating before next deploy attempt. The current `LibProdDeployV3.t.sol` tests assert `addr.codehash == EXPECTED_CODEHASH` against the pinned address constants, but those tests pass cleanly in this branch — which is suspicious given the bytecode shouldn't be deployable. Either:
[...]

PR #145, opened 2026-05-05
DO NOT MERGE — experiment: enable via_ir to measure size impact (#144)

thedavidmeister:
**DO NOT MERGE.** Experiment branch attached to #144 so the negative result has a PR-shaped record.

## TL;DR

`via_ir = true` makes `StoxReceiptVault` **larger**, not smaller. The contract that was already over EIP-170 (#90) gets +1,852 bytes more.

## Measurement

| Contract | Legacy (route 0) | IR (route 1) | Δ |
|---|---|---|---|
| StoxReceiptVault | 26,134 | **27,986** | **+1,852 (worse)** |
| TestStoxReceiptVault | 27,699 | 28,354 | +655 (worse) |
| StoxCorporateActionsFacet | 9,361 | 8,437 | -924 (better) |
| StoxWrappedTokenVault | 8,685 | 8,631 | -54 |
| StoxReceipt | 15,642 | 15,919 | +277 |

## Why

The IR pipeline aggressively monomorphises and inlines for runtime gas at high `optimizer_runs` (100k). Vault has deep inheritance + heavy inlining that the IR optimiser further inlines rather than collapses. With `optimizer_runs = 1` IR could plausibly invert (separate measurement; that's a different lever already listed in #90).

## Discard signal

#144 specified: "If size delta is <500 bytes, IR alone won't close the gap..." — actual delta is **+1,852 worse**, so even more conclusively a no-go on this lever.

## What to do instead

Per #90's remaining options:
- Move more logic to `StoxCorporateActionsFacet` (extract from vault).
- Lower `optimizer_runs` (deploy-size vs runtime-gas tradeoff).
- Lift OZ inheritance to library calls.

Closing this PR after merging is **not the plan**. It will be closed as a "tried, didn't work" experiment, and #144 is already closed with the same finding.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -5,6 +5,7 @@
 solc = "0.8.25"
 optimizer = true
 optimizer_runs = 100000
+via_ir = true
 
 evm_version = "cancun"
 


thedavidmeister:
Closing without merge per the PR body. Negative finding — IR worsens vault size by 1,852 bytes. Branch left in place; the PR record + #144 close comment together document the experiment for anyone scanning size-reduction options later.

PR #150, opened 2026-05-07
refactor: extract shared rebase-migration skeleton into RebaseMigratable (#94)

### LINKED ISSUE #94: Extract shared rebase-migration skeleton into a RebaseMigratable base contract
`StoxReceiptVault._migrateAccount` and `StoxReceipt._migrateHolderId` share an algorithmic skeleton: zero-address short-circuit, read cursor + stored balance, call the rebase walk, early-return if cursor unchanged, write cursor, conditionally write balance + emit event, post-migration hook. The differences are all plug-points (which storage lib, which rebase walk, which event signature, presence/absence of `LibTotalSupply.onAccountMigrated`), not structural.

Duplicating the skeleton makes any refactor of the migrate algorithm a two-file change. Miss the mirror edit and the two sides silently drift.

Proposed shape:

```solidity
abstract contract RebaseMigratable {
    function _readCursor(address account, uint256 id) internal view virtual returns (uint256);
    function _writeCursor(address account, uint256 id, uint256 cursor) internal virtual;
    function _readStoredBalance(address account, uint256 id) internal view virtual returns (uint256);
    function _writeStoredBalance(address account, uint256 id, uint256 balance) internal virtual;
    function _walkRebase(uint256 stored, uint256 cursor) internal view virtual returns (uint256, uint256);
    function _emitMigrated(address account, uint256 id, uint256 fromCursor, uint256 toCursor, uint256 oldBalance, uint256 newBalance) internal virtual;
    function _postMigrate(uint256 fromCursor, uint256 toCursor, uint256 oldBalance, uint256 newBalance) internal virtual;

    function _migrate(address account, uint256 id) internal {
        if (account == address(0)) return;
        uint256 currentCursor = _readCursor(account, id);
        uint256 storedBalance = _readStoredBalance(account, id);
        (uint256 newBalance, uint256 newCursor) = _walkRebase(storedBalance, currentCursor);
        if (newCursor == currentCursor) return;
        _writeCursor(account, id, newCursor);
        if (newBalance != storedBalance) {
            _writeStoredBalance(account, id, newBalance);
            _emitMigrated(account, id, currentCursor, newCursor, storedBalance, newBalance);
        }
        _postMigrate(currentCursor, newCursor, storedBalance, newBalance);
    }
}
```

Share-side passes `id=0` (ignored by its hooks). Receipt-side uses `id`. Share's `_postMigrate` calls `LibTotalSupply.onAccountMigrated`, receipt's is empty.

Caveats: widens the inheritance chain on both `StoxReceiptVault` and `StoxReceipt`; `id=0` is a slight tell on the share side. Trade for: one canonical place where the migrate algorithm lives, and drift prevention becomes structural rather than a review-time rule.

thedavidmeister:
Closes #94. **Up for review** — the trade-off (size + per-call cost vs structural drift prevention) wants a call before merging.

## What changed

Both `StoxReceiptVault._migrateAccount` and `StoxReceipt._migrateHolderId` ran the same 7-step orchestration:

1. Zero-address short-circuit
2. Read cursor + stored balance
3. Walk rebase from cursor
4. Early-return if cursor unchanged
5. Write new cursor
6. Conditional balance SSTORE + emit event
7. Post-migration hook

Differences were entirely plug-points (storage lib, rebase walk, event signature, presence/absence of `LibTotalSupply.onAccountMigrated`).

Extracted the orchestration into `src/abstract/RebaseMigratable.sol` with seven virtual hooks:
- `_readCursor` / `_writeCursor`
- `_readStoredBalance` / `_writeStoredBalance`
- `_walkRebase`
- `_emitMigrated`
- `_postMigrate`

`StoxReceiptVault` and `StoxReceipt` both inherit `RebaseMigratable` and implement the hooks. The algorithm shape now lives in one place; future migrate refactors edit one file instead of mirroring two.

## Trade-offs to weigh

**Vault bytecode size: +130 bytes.** 24,076 → 24,206. EIP-170 margin: +500 → +370. Still fits.

**Receipt per-call overhead.** Previously `_update` cached `_vault()` once and passed it through to every `_migrateHolderId` in the batch. Under the unified shape, `_walkRebase` does the `this.manager()` self-call per migration. Cost: one extra CALL opcode per `(account, id)` pair in the batch. For batch transfers this is real (multiplied by `ids.length * 2`). Single-id transfers and balance-of-batch reads aren't affected (they don't go through `_migrate`).

If batch overhead is unacceptable, alternatives:
- Use EIP-1153 transient storage to cache `_vault()` for the duration of the `_update` call (one TSTORE/TLOAD per migration vs one CALL).
- Keep an extra `_walkRebase(stored, cursor, vault)` overload on the receipt side and a separate batched entry point — but that re-introduces some duplication.

**`id` parameter on share side is always 0.** Minor smell. Share has no ERC-1155 token id concept; the override ignores `id` and the share-side `AccountMigrated` event drops the field. The unified `_migrate(account, id)` shape pays for the orchestration savings with this slight imprecision. Documented in `RebaseMigratable` NatSpec.

## Test plan
- [x] full suite green (513/513 pass) after pointer regen
- [x] vault still under EIP-170 (+370 margin)
- [ ] static + legal (CI)
- Reviewer judgement: is the +130 bytes / per-call CALL worth the drift prevention?

## Summary by CodeRabbit

* **Refactor**
  * Restructured rebase-migration framework with a new unified abstraction layer for standardized migration orchestration across contracts
  * Updated related contract implementations and test suites to integrate with the new architecture while preserving existing migration behavior


thedavidmeister:
🤖 ai:vetter
Reviewed 97b7da02c141986c656a86a2dfc67d81159a5238: design — closes #94 — RebaseMigratable extraction is behavior-preserving (verified: unconditional emit after cursor advance matches BOTH original sides, onAccountMigrated arg order kept, op order identical, zero-address short-circuit kept; correctly deviates from the issue's illustrative skeleton which guarded the emit); parks on the trade-off the PR itself flags for review: +130B vault bytecode (EIP-170 margin 500→370), per-pair _vault() CALL in receipt batch transfers vs once-per-_update caching, id=0 smell on share side — accept the cost or pick the transient-storage/overload alternative; sol-static red, cause not in the diff's semantics, producer to green after ruling
cost 712 — inheritance refactor over rebase accounting path

thedavidmeister:
👤 human
Ruled 97b7da02c141986c656a86a2dfc67d81159a5238: needs-work — The design question cannot be answered as posed: its premise describes a tree main no longer has. This branch is 355 commits behind main and still on git submodules, while main is on soldeer (soldeer.lock/remappings.txt/.soldeerignore, libs = ['dependencies']); the +130B / "EIP-170 margin 500→370" numbers were computed against this branch's stale foundry.toml at optimizer_runs = 24483, where main runs optimizer_runs = 2000 with the vault at 24,037 and 539 bytes spare; and main has renamed the two functions this diff rewrites (_migrateHolderId → migrateHolderId, _vault → getVault). CONFLICTING is the same fact showing up mechanically.

thedavidmeister:
Rework note @97b7da02c141986c656a86a2dfc67d81159a5238: Merge main in (never rebase) and port the RebaseMigratable extraction forward: main is on soldeer with no submodules, and the functions this PR rewrites have been renamed — `_migrateHolderId` → `migrateHolderId`, `_vault()` → `getVault()`. Then recompute the size cost against main's real config (`optimizer_runs = 2000`, StoxReceiptVault at 24,037 with 539 bytes spare) via `forge build --sizes`, and re-raise the trade-off only with those numbers. Do NOT change `optimizer_runs` to buy margin: main's foundry.toml documents that value as a coordinated-release event, not a drive-by — it moves every contract's deterministic Zoltu address, and StoxReceiptVault pins the corporate-actions facet while ST0xOrchestrator's version lock pins the vault + receipt implementations, so all three must deploy in lockstep. Before re-raising the per-pair `_vault()` CALL trade-off at all, check whether it still exists: that CALL is forced only because rain.vats declares `manager()` `external` (Receipt.sol:175) while `getStorageReceipt()` is `private` (Receipt.sol:47), so an upstream internal accessor removes the cost entirely and makes the question moot — see the rain.vats issue filed alongside this ruling, and prefer waiting on it to adding EIP-1153 transient caching or a `_walkRebase` overload. `id=0` on the share side and the widened inheritance chain are NOT reopened: issue #94 weighed and accepted both explicitly before the work started. Also fix `StoxReceipt._vault()`'s NatSpec, which still claims the cost is "one CALL opcode per `_update`, amortized across every `(account, id)` pair" — the extraction deleted exactly that amortization. Then green sol-static.

PR #222, opened 2026-07-01
Add ST0xOrchestrator: per-token mint/burn proxy

hardyjosh:
**Redesigned as a singleton** per the 2026-07-02 design meeting + David's review. One instance serves every token; all per-token state is keyed by the receipt-vault address.

## Interface
- `mint(address token, address to, uint256 amount, bytes data)` — MINT_ROLE. Sends shares to `to`, which must authorise the mint (see below). Receipt kept by the orchestrator.
- `burn(address token, address from, uint256 amount, bytes data)` — BURN_ROLE. Per-token burn-walk; reverts `InsufficientReceipts` if held receipts can't cover the amount (never mints to cover a shortfall — manual recovery).
- `data` carries `receiptInformation` forwarded to `vault.mint`/`redeem` for the on-chain audit trail.

## Mint recipient authorisation
Stops a compromised MINT key directing freshly minted shares anywhere. Every mint carries `to`'s authorisation of `(token, to, amount, nonce)` as a versioned `MintAuthV1` struct: an EIP-712 signature (via `SignatureChecker` — EOA ECDSA **or** contract EIP-1271) **or**, when no signature is supplied, an `IMintRecipient.authorizeMint` callback (ERC-1155-receiver style — lets a keyless contract like the atomic bridge gate on its own on-chain intent). Replay protection is namespaced by recipient — `(to, nonce)` is single-use regardless of token or amount, so nobody can consume another recipient's nonce. The minter's `receiptInformation` audit payload is a separate parameter, never part of the recipient's authorisation. The full external interface (functions, events, errors, `Digest` alias type, `MintAuthV1`) lives in `src/interface/IST0xOrchestratorV1.sol` for other contracts to import.

## Roles
- `MINT_ROLE`, `BURN_ROLE` — split (not a combined role).
- `EMERGENCY_ROLE` — setBurnIndex / withdrawReceipt / withdrawShares / sweepERC1155. Separate from mint/burn so the recovery key can never mint.
- `DEFAULT_ADMIN_ROLE` — administers roles only; performs no operations.

## Vault-logic version lock
`initialize` and `mint`/`burn` revert unless the shared production vault + receipt beacons still point at the `LibProdDeployV4` implementations this orchestrator was built against. A vault upgrade halts the orchestrator until it is upgraded in lockstep (mirrors the vault baking the corporate-actions facet into its bytecode).

## Burn walk (per-token)
Walks a per-token `nextBurnReceiptId` pointer over the orchestrator's own rebased receipt balances; burning more than held reverts `InsufficientReceipts(token, shortfall)` — a shortfall is an anomaly to recover manually (transfer receipts in / `setBurnIndex`), never papered over by minting. The ERC-1155 receiver hooks self-maintain the pointer: a genuine production receipt (defensive `manager()`/`receipt()` round-trip) arriving below the pointer lowers it (`BurnIndexLowered`), so transferred-in receipts are always burnable with no manual `setBurnIndex`. A fresh token's pointer starts at 0.

## Tests — 75 (all green)
- 58 unit+fuzz (all three auth modes incl. EIP-1271 + callback, `(to, nonce)` replay matrix, role split + cross-role negatives, guard pass/fail incl. at-initialize, per-token walks, insufficient-receipts reverts, receiver-hook lower/no-op/spoof/batch, sweeps).
- 11 real-V4-deploy integration tests, one workflow per file under `test/src/concrete/integration/` (guard against real beacons, ECDSA + callback mint, burn happy path, rebase 3:1 split exact-consume, 1:3 split truncation-dust revert + partial-burn recovery, guard-halt-on-beacon-upgrade, cert lapse, receipt-transfer-in-lowers-pointer ops bootstrap).
- 9 deployer.

## Review
Ran a 2-lane security + correctness panel: no critical/high/medium. Applied: `ReentrancyGuardTransient` on all state-changing entrypoints, guard shared-beacon scope documented, distinct `ForeignERC1155Swept` event.

## Addresses David's review comments
- **Combined role** → split into `MINT_ROLE` + `BURN_ROLE`.
- **DEFAULT_ADMIN overloaded** → dedicated `EMERGENCY_ROLE`; admin administers roles only.
- **Weak vault guard** → real runtime version lock against the shared beacon impls.
- **mint-to-other offchain behaviour** → recipient-authorisation requirement.
- **View fn ordering** → views moved below state-changing functions.

**Do not merge yet** — pairs with #223 (deploy wiring). Deploy + permissions scripts follow.

## Summary by CodeRabbit

* **New Features**
  * Introduced **ST0xOrchestrator** singleton with role-gated minting, burning, and emergency controls, using EIP-712/callback recipient authorization with replay protection.
  * Added production receipt “version lock” guarding against unexpected vault logic.
  * Added ERC-1155 receipt handling with sequential burn-pointer mechanics and emergency sweeping/withdrawals.
  * Added **ST0xOrchestratorBeaconSetDeployer** for permissionless proxy deployment of new orchestrator instances.
* **Tests**
  * Expanded unit/integration coverage for authorization, signature/callback flows, replay/reentrancy, burn-pointer updates, emergency paths, and beacon upgrade behavior.
* **Documentation**
  * Updated changelog and release artifacts to include the orchestrator components.


hardyjosh:
* **#224** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#223** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/223?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#222** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/222?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/222?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,181 @@
...
+
+    /// @notice The next receipt id the burn walk will consider. Burns scan
+    /// forward from this index, skipping zero-balance ids, until the
+    /// requested amount is satisfied. Bumped past fully-consumed ids.
+    uint256 public nextBurnReceiptId;
+
+    event Minted(address indexed caller, address indexed to, uint256 amount);
+    event Burned(
+        address indexed caller, address indexed from, uint256 amount, uint256 firstReceiptId, uint256 lastReceiptId
+    );
```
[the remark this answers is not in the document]
hardyjosh: Fixed at 8c7ee9d — renamed the event field to `nextBurnReceiptIdAfter` with NatSpec calling it out as the walk pointer (not the last-touched id).

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,181 @@
...
+    event BurnIndexSet(uint256 oldIndex, uint256 newIndex);
+    event ReceiptsWithdrawn(address indexed to, uint256 indexed id, uint256 amount);
+    event SharesWithdrawn(address indexed to, uint256 amount);
+
+    constructor(OffchainAssetReceiptVault vault_, address initialAdmin) {
+        vault = vault_;
+        receipt = IERC1155(address(vault_.receipt()));
+        _grantRole(DEFAULT_ADMIN_ROLE, initialAdmin);
+    }
```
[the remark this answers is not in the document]
hardyjosh: Addressed. Current head has a parameterless constructor (`_disableInitializers()`) and an `initialize(vault, owner)` external function — the Initializable pattern rather than ICloneableV2, because per-token instances are `BeaconProxy` clones minted by `ST0xOrchestratorBeaconSetDeployer`, not Rain `CloneFactory` clones. Deterministic Zoltu deploy works fine for the impl (no constructor args). See the tests in `test/src/concrete/ST0xOrchestrator.t.sol::testConstructorDisablesInitializers` and the deployer tests.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,181 @@
...
+
+    function supportsInterface(bytes4 interfaceId) public view override(AccessControl, IERC165) returns (bool) {
+        return interfaceId == type(IERC1155Receiver).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    /// @dev `ReceiptVault.mint` is payable and refunds any ETH the vault
+    /// holds to `msg.sender` (this orchestrator) via `Address.sendValue`.
+    /// Accept it and let it sit; admin can sweep if it ever matters.
+    receive() external payable {}
```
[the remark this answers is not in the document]
hardyjosh: Not adding a sweep — the orchestrator is not supposed to handle ETH. The refund path via `vault.mint`s `Address.sendValue` is always zero-value in practice (the vault has no ETH balance under normal operation). Rewrote the NatSpec to say so explicitly instead of implying an admin sweep exists. If ETH ever ends up here it means something misbehaved upstream; adding a sweep would just paper over the anomaly.

src/concrete/deploy/ST0xOrchestratorBeaconSetDeployer.sol
```diff
@@ -0,0 +1,81 @@
...
+
+    /// The beacon every deployed orchestrator proxy points at.
+    IBeacon public immutable iOrchestratorBeacon;
+
+    constructor(ST0xOrchestratorBeaconSetDeployerConfig memory config) {
+        if (config.initialOwner == address(0)) revert ZeroInitialOwner();
+        if (config.initialOrchestratorImplementation == address(0)) revert ZeroOrchestratorImplementation();
+
+        iOrchestratorBeacon = new UpgradeableBeacon(config.initialOrchestratorImplementation, config.initialOwner);
+    }
```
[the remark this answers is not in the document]
hardyjosh: This is the base class — the parameterless subclass with hardcoded config is `StoxST0xOrchestratorBeaconSetDeployer` in PR #223, mirroring the `OffchainAssetReceiptVaultBeaconSetDeployer` / `StoxOffchainAssetReceiptVaultBeaconSetDeployer` pair in rain-vats. The base with a config-in-constructor stays here for testability + reuse; the Stox subclass exists purely to be Zoltu-deployable.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,246 @@
...
+    /// @param owner Address granted `DEFAULT_ADMIN_ROLE`. In production this
+    /// is the owner multisig — it holds unilateral authority over
+    /// `setBurnIndex`, `withdrawReceipt`, `withdrawShares`, and
+    /// `MINT_BURN_ROLE` grants and revocations.
+    function initialize(OffchainAssetReceiptVault vault_, address owner) external initializer {
+        __AccessControl_init();
+        MainStorage storage $ = _main();
+        $.vault = vault_;
+        $.receipt = IERC1155(address(vault_.receipt()));
+        _grantRole(DEFAULT_ADMIN_ROLE, owner);
```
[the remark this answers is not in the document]
hardyjosh: Fixed. `initialize` now reverts `ZeroVault` / `ZeroOwner` / `ZeroReceipt` on any of those. Tests added at `test/src/concrete/ST0xOrchestrator.t.sol::testInitializeZeroVaultReverts`, `testInitializeZeroOwnerReverts`, `testInitializeZeroReceiptReverts`.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,246 @@
...
+        return IERC1155Receiver.onERC1155Received.selector;
+    }
+
+    function onERC1155BatchReceived(address, address, uint256[] calldata, uint256[] calldata, bytes calldata)
+        external
+        pure
+        returns (bytes4)
+    {
+        return IERC1155Receiver.onERC1155BatchReceived.selector;
+    }
```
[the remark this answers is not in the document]
hardyjosh: Fixed. `onERC1155Received` / `onERC1155BatchReceived` both revert `UnrecognisedERC1155Source(msg.sender)` when called by anything other than the configured receipt token. Tests: `testFuzzOnERC1155ReceivedFromReceipt` / `testFuzzOnERC1155BatchReceivedFromReceipt` for the accept path, `testFuzzOnERC1155ReceivedRejectsOthers` / `testFuzzOnERC1155BatchReceivedRejectsOthers` for the rejection.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,270 @@
...
+    /// @notice Mint `amount` rebased tStocks to `to` via `vault.mint`. The
+    /// fresh ERC-1155 receipt is minted to this orchestrator (kept here
+    /// permanently); the share balance is forwarded to `to`.
+    function mint(address to, uint256 amount) external onlyRole(MINT_BURN_ROLE) {
+        MainStorage storage $ = _main();
+        $.vault.mint(amount, address(this), 0, "");
+        if (to != address(this)) {
+            IERC20(address($.vault)).safeTransfer(to, amount);
+        }
+        emit Minted(msg.sender, to, amount);
```
thedavidmeister: slither doesn't complain about this?

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,270 @@
...
+        if (receipt_ == address(0)) revert ZeroReceipt();
+
+        __AccessControl_init();
+        MainStorage storage $ = _main();
+        $.vault = vault_;
+        $.receipt = IERC1155(receipt_);
+        _grantRole(DEFAULT_ADMIN_ROLE, owner);
+    }
+
+    // ------------------------------------------------------------------ //
```
thedavidmeister, line 130: maybe better to put these lower as earlier declared functions cost less gas and i guess these are largely for offchain use

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,270 @@
...
+/// each initialised with its own `(vault, owner)`.
+///
+/// **No bootstrap on `initialize`.** Receipts are transferred in after
+/// deploy (e.g. from a legacy issuer EOA holding pre-existing receipts).
+/// After the transfer, an admin calls `setBurnIndex` to point the pointer at
+/// the first id with non-zero balance.
+contract ST0xOrchestrator is Initializable, AccessControlUpgradeable, IERC1155Receiver {
+    using SafeERC20 for IERC20;
+
+    bytes32 public constant MINT_BURN_ROLE = keccak256("MINT_BURN");
```
thedavidmeister: why a combined role? that's strictly less flexible than having two roles and assigning both if/as needed

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,270 @@
...
+
+    /// @notice Initialise a per-token orchestrator clone.
+    /// @param vault_ The receipt vault this instance mints and burns against.
+    /// Fixed for the lifetime of the clone.
+    /// @param owner Address granted `DEFAULT_ADMIN_ROLE`. In production this
+    /// is the owner multisig — it holds unilateral authority over
+    /// `setBurnIndex`, `withdrawReceipt`, `withdrawShares`, and
+    /// `MINT_BURN_ROLE` grants and revocations.
+    function initialize(OffchainAssetReceiptVault vault_, address owner) external initializer {
+        if (address(vault_) == address(0)) revert ZeroVault();
```
thedavidmeister: this feels like a very weak guard considering how much of the contract logic expects the vault to work in a very specific way, why not actually check the vault logic is what we expect at init time? is it worth also locking the vault logic to expectations across upgrades at runtime even?

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,270 @@
...
+            IERC20(address(vault_)).safeTransferFrom(from, address(this), amount);
+        }
+
+        uint256 idx = $.nextBurnReceiptId;
+        uint256 startIdx = idx;
+        uint256 remaining = amount;
+        // Cap the walk at the vault's current highwater. Beyond this id no
+        // receipt can possibly exist, so further `balanceOf` calls would
+        // walk forever. When we hit the cap and still need to burn, mint
+        // a fresh receipt-backed batch to cover the shortfall — net zero
```
thedavidmeister: not sure this forces the desired offchain behaviour that we want, even if it nets out onchain

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done — the view functions (nextBurnReceiptId, tokenSeeded, mintAuthUsed, mintAuthDigest, vaultLogicIsExpected) are now below the state-changing entrypoints.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Split into separate MINT_ROLE + BURN_ROLE (and a distinct EMERGENCY_ROLE for setBurnIndex/withdraw*). Assign both if/as needed.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Reworked into a real runtime version lock: `onlyExpectedVaultLogic` reads the shared production vault + receipt beacons and reverts mint/burn/advanceBurnIndex unless they still point at the LibProdDeployV4 implementations the orchestrator was built against. A vault upgrade halts the orchestrator until it is upgraded in lockstep. Integration test `testGuardHaltsAfterVaultBeaconUpgrade` covers the halt. (Scope note in the modifier natspec: it checks the shared beacons, not the specific token arg — a token not on the shared set has no role grant so mint/burn revert at the authoriser anyway.)

src/concrete/ST0xOrchestrator.sol
hardyjosh: Redesigned: mint now requires the recipient to authorise (token,to,amount,nonce) — an EIP-712 signature (EOA ECDSA or contract EIP-1271 via SignatureChecker) or an IMintRecipient.authorizeMint callback. A compromised MINT key cannot direct shares to a destination that has not signed/callback-approved. The callback path lets a keyless contract (e.g. the atomic bridge) gate on its own on-chain intent.

src/concrete/ST0xOrchestrator.sol
hardyjosh: The receiver hooks are now pure no-ops (accept-all) with a generic sweepERC1155 recovery, so the earlier slither concern is moot. The redesign added ReentrancyGuardTransient on all state-changing entrypoints as defence-in-depth (the mint recipient callback runs before vault.mint).

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,481 @@
...
+///
+/// **Vault-logic version lock.** So much of the burn/mint logic depends on
+/// the exact behaviour of the current receipt-vault implementation that
+/// `mint`/`burn` refuse to run unless the production vault + receipt beacons
+/// still point at the implementations this orchestrator was built against
+/// (`LibProdDeployV4`). If the vault is upgraded, the orchestrator halts
+/// until its own implementation is upgraded in lockstep. This mirrors the
+/// vault baking the corporate-actions facet address into its own bytecode.
+///
+/// **Burn walk.** `burn` walks a per-token `nextBurnReceiptId` pointer over
```
hardyjosh, line 62: we don't want to do this. if the orchestrator doesn't have the receipts to burn that's a revert. we'd have to recover that manually.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,481 @@
...
+        address indexed from,
+        uint256 amount,
+        uint256 firstReceiptId,
+        uint256 nextBurnReceiptIdAfter
+    );
+    /// @notice The burn walk exhausted the orchestrator's held receipts for
+    /// `token` and minted a fresh receipt-backed batch to cover the shortfall
+    /// (interest-accrual case, or rebase-truncation dust after fractional
+    /// splits, or a mis-set pointer — monitor accordingly).
+    event BurnShortfallMinted(address indexed token, uint256 amount, uint256 receiptId);
```
hardyjosh: as above - don't want this

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done at 10a1491 — mint-on-demand removed. The burn walk now reverts `InsufficientReceipts(token, shortfall)` whenever it crosses `highwaterId` with anything unburned, so the whole burn reverts and recovery is manual (transfer receipts in / `setBurnIndex`). Knock-on in the atomic bridge: `liquidate()` rounded the Morpho share slice up, which could demand one wei more tStock than was ever minted — now rounds down so the pulled assets can never exceed the request (the wei stays as debt).

src/concrete/ST0xOrchestrator.sol
hardyjosh: Removed — `BurnShortfallMinted` is gone along with the mint-on-demand path. The fractional-split integration test now asserts the truncation-dust burn reverts and demonstrates the partial-burn + manual-recovery path instead.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+    constructor() {
+        _disableInitializers();
+    }
+
+    /// @notice Initialise the singleton. Grants `DEFAULT_ADMIN_ROLE` to
+    /// `owner` (the owner multisig). Operational roles (`MINT_ROLE`,
+    /// `BURN_ROLE`, `EMERGENCY_ROLE`) are granted separately by the admin.
+    /// @param owner Address granted `DEFAULT_ADMIN_ROLE` — the role admin
+    /// only; it performs no mint/burn/recovery operations itself.
+    function initialize(address owner) external initializer {
```
hardyjosh, line 122: add onlyExpectedVaultLogic modifierhere

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+
+    /// @dev Decode `data`, verify + consume the recipient's single-use mint
+    /// authorisation, and return `(nonce, receiptInformation)`. Split out of
+    /// `mint` to keep that frame within stack limits.
+    function _consumeMintAuth(address token, address to, uint256 amount, bytes calldata data)
+        internal
+        returns (bytes32 nonce, bytes memory receiptInformation)
+    {
+        bytes memory signature;
+        (signature, nonce, receiptInformation) = abi.decode(data, (bytes, bytes32, bytes));
```
hardyjosh: this shape should be a versioned struct that can also be consumed by other contracts. actually the whole orchestrator interface should be a in a separate file along with this.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+        bytes32 digest = _mintAuthDigest(token, to, amount, nonce);
+        MainStorage storage $ = _main();
+        if ($.usedMintAuth[digest]) revert MintAuthReplayed(digest);
+        $.usedMintAuth[digest] = true;
```
hardyjosh: we want the replay protection to be stronger actually - just hash the `to` and the `nonce`. it's up the caller to make sure that nonce's are unique, regardless of the token or amount.  
  
digest is probably the wrong naming for where this is being used. it's just a replay id.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+    /// @dev Decode `data`, verify + consume the recipient's single-use mint
+    /// authorisation, and return `(nonce, receiptInformation)`. Split out of
+    /// `mint` to keep that frame within stack limits.
+    function _consumeMintAuth(address token, address to, uint256 amount, bytes calldata data)
+        internal
+        returns (bytes32 nonce, bytes memory receiptInformation)
+    {
+        bytes memory signature;
+        (signature, nonce, receiptInformation) = abi.decode(data, (bytes, bytes32, bytes));
+        bytes32 digest = _mintAuthDigest(token, to, amount, nonce);
```
hardyjosh: _mintAuthDigest should return an alias type instead of bytes32 so it can't be misused, and also _verifyRecipientAuth can't accept a bytes32 that wasn't produced explicitly as a digest

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+    /// called back. `receiptInformation` is forwarded to `vault.mint` for the
+    /// on-chain audit trail.
+    function mint(address token, address to, uint256 amount, bytes calldata data)
+        external
+        onlyRole(MINT_ROLE)
+        onlyExpectedVaultLogic
+        nonReentrant
+    {
+        if (amount == 0) revert ZeroAmount();
+        (bytes32 nonce, bytes memory receiptInformation) = _consumeMintAuth(token, to, amount, data);
```
hardyjosh: the receiptInformation is not part of the mint auth, so it shouldn't be the responsibility of _consumeMintAuth to decode the data param.  
  
and more broadly, mint auth is the responsibility of the receiver not the minter.  
  
we need a step first to pull out the receiptInformation and then pass only the mint auth to the consume mint auth function  
  
this would be fixed if we had a struct which was the expected shape of the data field. see the other comment about the interface.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+    /// history. Bootstraps of transferred-in receipts at lower ids use an
+    /// `EMERGENCY_ROLE` `setBurnIndex` afterwards.
+    function _seedToken(address token) internal {
+        MainStorage storage $ = _main();
+        if ($.tokenSeeded[token]) return;
+        $.tokenSeeded[token] = true;
+        uint256 seeded = OffchainAssetReceiptVault(payable(token)).highwaterId() + 1;
+        $.nextBurnReceiptId[token] = seeded;
+        emit TokenSeeded(token, seeded);
+    }
```
hardyjosh: not needed - for a token that already has receipts operationally the process is simply:  
1\. transfer in any existing receipts for that receiptvault  
2\. admin set the burn index to the lowest id that was transferred in, OR - see below comment on 454  
3\. grant the orchestrator the deposit and withdraw roles on the receiptvault

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+    //                          ERC-1155 receiver                         //
+    // ------------------------------------------------------------------ //
+
+    /// @dev A singleton cannot cheaply identify every legitimate receipt
+    /// token at callback time (each token has its own receipt, and a
+    /// `BeaconProxy`'s beacon is not externally readable), so the hooks
+    /// accept all senders. This is safe: the orchestrator only credits its
+    /// burn walk against receipts its OWN `vault.mint` created, and any
+    /// foreign ERC-1155 that lands here is recoverable via `sweepERC1155`.
+    function onERC1155Received(address, address, uint256, uint256, bytes calldata) external pure returns (bytes4) {
```
hardyjosh: here we can check if the id of the received receipt is lower than our current burn index, and if it is, lower it. this means we'll potentially never need to use the admin setBurnIndex. but we can keep it just in case

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+    /// @param token The `OffchainAssetReceiptVault` to mint.
+    /// @param to Recipient of the shares. Must authorise via `data`.
+    /// @param amount Rebased tStock units to mint.
+    /// @param data `abi.encode(bytes signature, bytes32 nonce, bytes
+    /// receiptInformation)`. If `signature` is non-empty it is checked
+    /// against `to` (ECDSA or EIP-1271) over the EIP-712 digest of
+    /// `(token, to, amount, nonce)`; if empty, `IMintRecipient(to)` is
+    /// called back. `receiptInformation` is forwarded to `vault.mint` for the
+    /// on-chain audit trail.
+    function mint(address token, address to, uint256 amount, bytes calldata data)
```
hardyjosh: we're rethinking this - data should be split into the mint auth digest/sig and receiptInformation can be a separate data field that gets forwarded through. there isn't much point having a data field because the callers need to know the encoding and any future upgrade would probably just get a new versioned contract interface anyway (mint2)

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+        _verifyRecipientAuth(to, digest, signature);
+    }
+
+    /// @notice Burn `amount` rebased tStocks of `token` from `from`. Pulls
+    /// the shares (if not already held), then walks the per-token pointer.
+    /// Reverts `InsufficientReceipts` if the orchestrator's held receipts
+    /// cannot cover `amount` — recover manually, never by minting.
+    /// @param data `receiptInformation` forwarded to `vault.redeem` for the
+    /// audit trail — e.g. a tag marking a debt-repay burn.
+    function burn(address token, address from, uint256 amount, bytes calldata data)
```
hardyjosh: let's just call this what it is - burn info

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+    /// @notice Walk `token`'s burn pointer forward across zero-balance ids
+    /// WITHOUT burning, persisting progress. Permissionless: it can only skip
+    /// ids the orchestrator holds no balance at, stops at the first non-zero
+    /// balance, and never passes `highwaterId + 1` — so no caller can strand
+    /// receipts held at call time or misposition the pointer. Lets keepers
+    /// cross a long zero-balance gap incrementally (bounded by `maxIds`)
+    /// instead of forcing a single burn to cross it and risk out-of-gas.
+    /// Receipts arriving later at already-skipped ids are inert until an
+    /// `EMERGENCY_ROLE` `setBurnIndex`, as with `burn`.
+    function advanceBurnIndex(address token, uint256 maxIds)
```
hardyjosh: we don't need this - it's solving a problem we don't really have

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+    /// after receipts are transferred in, or recover from a mis-set pointer.
+    /// @dev O(gap) hazard, both directions: set too LOW and the next `burn`
+    /// pays one external rebased `balanceOf` per id to cross the gap in a
+    /// single tx (mitigate with `advanceBurnIndex`); set too HIGH and held
+    /// receipts behind the pointer are stranded, so burns revert
+    /// `InsufficientReceipts` once the receipts ahead of it are exhausted.
+    /// Set at (or just below) the first id with non-zero balance.
+    function setBurnIndex(address token, uint256 newIndex) external onlyRole(EMERGENCY_ROLE) nonReentrant {
+        MainStorage storage $ = _main();
+        $.tokenSeeded[token] = true;
```
hardyjosh: we don't need this

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,480 @@
...
+        address indexed caller,
+        address indexed token,
+        address indexed from,
+        uint256 amount,
+        uint256 firstReceiptId,
+        uint256 nextBurnReceiptIdAfter
+    );
+    event BurnIndexSet(address indexed token, uint256 oldIndex, uint256 newIndex);
+    event BurnIndexAdvanced(address indexed caller, address indexed token, uint256 oldIndex, uint256 newIndex);
+    event TokenSeeded(address indexed token, uint256 pointer);
```
hardyjosh: not needed

test/src/concrete/ST0xOrchestrator.integration.t.sol
hardyjosh, line 1: let's make each integration test/workflow a separate file in an integration folder

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done at 841fb0b — `initialize` now calls `_checkVaultLogic()` (the modifier body, refactored into a shared internal), so an orchestrator can never be initialised against vault logic it wasn't built for. Unit tests cover both mismatch legs at init time.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done — new `src/interface/IST0xOrchestratorV1.sol` holds the full external interface (all functions, events, errors), the versioned `MintAuthV1` struct, and the `Digest` type. The concrete contract inherits it; other contracts (e.g. the atomic bridge) import the interface instead of the concrete.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done — replay is now a nested `(to => nonce => used)` mapping: a recipient's nonce is single-use regardless of token or amount, and namespacing by `to` means nobody can consume another recipient's nonce from the mempool. `NonceReplayed(to, nonce)` replaces `MintAuthReplayed(digest)`, `nonceUsed(to, nonce)` replaces `mintAuthUsed(digest)`. The digest is no longer used for replay at all — it's only the thing that gets signed/verified. Tests cover same-nonce-different-amount and same-nonce-different-token reverting, and same-nonce-different-recipient succeeding.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done — `type Digest is bytes32` in the interface file. `mintAuthDigest`/`_mintAuthDigest` return `Digest`, `_verifyRecipientAuth` and `IMintRecipient.authorizeMint` take `Digest`. UDVTs erase to their underlying type in the ABI, so `authorizeMint`'s selector is unchanged. The compiler now rejects any bytes32 that wasn't explicitly produced as a digest.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done — `receiptInformation` is a separate top-level `mint` parameter (the MINTER's audit payload), and `_consumeMintAuth` takes only the `MintAuthV1` struct: it consumes the `(to, nonce)` replay slot and verifies the recipient's authorisation, nothing else. Responsibilities are now split exactly along the minter/recipient line.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done — `mint(token, to, amount, MintAuthV1 auth, bytes receiptInformation)`; the generic data blob is gone. `MintAuthV1` is versioned in the interface with a note that a breaking change ships as `MintAuthV2` + a new entrypoint, so callers break loudly at the ABI rather than silently mis-decoding.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done — renamed to `burnInfo` throughout (param, walk, NatSpec).

src/concrete/ST0xOrchestrator.sol
hardyjosh: Removed — `_seedToken`, the `tokenSeeded` mapping/view, and the `TokenSeeded` event are all gone. A fresh token's pointer starts at 0 and the ops process you describe covers existing-receipt tokens (with step 2 now usually automatic — see the receiver-hook change).

src/concrete/ST0xOrchestrator.sol
hardyjosh: Done — `onERC1155Received`/`onERC1155BatchReceived` now check whether the sender is a genuine production receipt (its `manager()` vault round-trips `vault.receipt() == msg.sender`, both as defensive staticcalls so a foreign/malicious 1155 can never revert the transfer or spoof a pointer — spoofing would require controlling the vault itself) and lower the burn pointer when a receipt arrives below it, emitting `BurnIndexLowered`. `setBurnIndex` stays as the EMERGENCY manual override. New integration workflow `ReceiptTransferInLowersPointer` covers the ops bootstrap end-to-end: transfer receipts in → pointer lowers automatically → burn succeeds with no admin call.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Removed — `advanceBurnIndex` and its event are gone; zero-balance gaps are crossed inside burns. Runtime size dropped to 12,546 B.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Removed along with the rest of `tokenSeeded` — `setBurnIndex` no longer writes it.

src/concrete/ST0xOrchestrator.sol
hardyjosh: Removed (the `TokenSeeded` event, with the mapping and view).

test/src/concrete/ST0xOrchestrator.integration.t.sol
hardyjosh: Done — the monolithic integration file is replaced by `test/src/concrete/integration/` with an abstract `OrchestratorIntegrationTest` base (shared real-V4 setUp + helpers) and one file per workflow: guard-passes, mint-with-ECDSA, mint-with-callback, burn-happy-path, burn-after-forward-split, burn-fractional-split-truncation-dust, guard-halts-after-beacon-upgrade, certification-lapse, and a new receipt-transfer-in-lowers-pointer workflow. 11 tests across 9 files, all green.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,481 @@
...
+    // ------------------------------------------------------------------ //
+    //                           Emergency sweeps                         //
+    // ------------------------------------------------------------------ //
+
+    /// @notice `EMERGENCY_ROLE` escape hatch: pull a specific receipt out.
+    function withdrawReceipt(address token, uint256 id, uint256 amount, address to) external onlyRole(EMERGENCY_ROLE) {
+        IERC1155(address(OffchainAssetReceiptVault(payable(token)).receipt()))
+            .safeTransferFrom(address(this), to, id, amount, "");
+        emit ReceiptsWithdrawn(token, to, id, amount);
+    }
```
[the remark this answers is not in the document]
hardyjosh: Fixed at c453710 — `withdrawReceipt` now carries `nonReentrant` like its siblings. EMERGENCY-gated so this is defence-in-depth, but there was no reason for the asymmetry.

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,270 @@
...
+        return IERC1155Receiver.onERC1155Received.selector;
+    }
+
+    function onERC1155BatchReceived(address, address, uint256[] calldata, uint256[] calldata, bytes calldata)
+        external
+        view
+        returns (bytes4)
+    {
+        if (msg.sender != address(_main().receipt)) revert UnrecognisedERC1155Source(msg.sender);
+        return IERC1155Receiver.onERC1155BatchReceived.selector;
```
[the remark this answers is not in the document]
hardyjosh: Not valid — `override` has not been required when implementing a function declared only in an inherited interface since Solidity 0.8.8 (it is only needed when multiple bases declare the same function). The contract compiles and the full suite is green on 0.8.25 with the hooks as written; they were also reworked in the latest push (they now maintain the burn pointer). Resolving.

thedavidmeister:
@coderabbitai resume

thedavidmeister:
@coderabbitai review

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,438 @@
...
+                continue;
+            }
+            uint256 take = remaining < bal ? remaining : bal;
+            // Share ratio is 1:1 by construction; anything else is the vault
+            // misbehaving and must halt the burn.
+            // slither-disable-next-line calls-loop
+            uint256 assets = vault.redeem(take, address(this), address(this), idx, burnInfo);
+            if (assets != take) revert VaultAmountMismatch(take, assets);
+            remaining -= take;
+            if (take == bal) {
```
thedavidmeister, line 252: seems redundant here due to 239

src/concrete/ST0xOrchestrator.sol
```diff
@@ -0,0 +1,438 @@
...
+    /// amount still unburned — the orchestrator never mints to cover a
+    /// shortfall. Persists and returns the final pointer. Split out of
+    /// `burn`, and `burnInfo` taken as `memory`, to keep both frames within
+    /// stack limits.
+    // Pointer write after external calls is safe: every caller holds the
+    // ReentrancyGuardTransient lock for the whole entrypoint.
+    // slither-disable-next-line reentrancy-no-eth
+    function _burnWalk(address token, uint256 remaining, bytes memory burnInfo) internal returns (uint256 idx) {
+        OffchainAssetReceiptVault vault = OffchainAssetReceiptVault(payable(token));
+        IERC1155 receipt_ = IERC1155(address(vault.receipt()));
```
thedavidmeister: don't use _ for local names

thedavidmeister: Renamed `receipt_` → `vaultReceipt` in `_burnWalk` (aff859b). Bytecode-neutral — the orchestrator pin is unchanged (`0x77ab9b240caC5F37a5D4d51651936ea1d61DF1A2`), so the on-chain deploy stands.

PR #229, opened 2026-07-06
feat(deploy): Ethereum V4 authoriser clone — chain-parametric deploy script + per-chain pin selector

hardyjosh:
Deploys the V4 authoriser clone on Ethereum by running the **same** script
that deployed it on Base, adapted only to pick the active chain's pin.

- **Deploy script** (`20260619-deploy-v4-authoriser-clone`) — already
  chain-agnostic (it reads the impl, Safe, and role grants from libs), so the
  only change is swapping the single-chain clone lookup for a per-chain one
  (`LibProdAuthoriserClones.cloneForChainId(block.chainid)`). Deploy on
  Ethereum by running it with `--rpc-url ethereum`; on Base it self-guards
  (`V4AuthoriserClonePinAlreadyHydrated`) because the Base pin is set.

- **Per-chain clone selection** — `activeChainClonePin()` in the deploy script
  dispatches on `block.chainid` to the active chain's `LibProdDeployV4` pin.
  It is a SELECTOR, not a second copy of the addresses: Base resolves the
  generated `STOX_PROD_AUTHORISER_V4_CLONE` pin (`0x315b…`) and Ethereum
  resolves `STOX_PROD_AUTHORISER_V4_CLONE_ETHEREUM`, an `address(0)`
  placeholder until the post-deploy hydrate PR fills it (the clone is a
  nonce-based `CloneFactory.clone`, so its address cannot be known ahead of
  the broadcast). An unpinned chain reverts
  `V4AuthoriserCloneUnsupportedChain` rather than falling back to another
  chain's clone. The chain ids come from `LibSafeInvariants.BASE_CHAIN_ID` /
  `ETHEREUM_CHAIN_ID`, the same constants the sibling per-chain selectors use.
  The EIP-1167 codehash is deterministic and shared, so it stays on the
  existing generated pin. **No hand-written change to `src/generated`** — the
  new Ethereum constant is emitted by `BuildPointers`, so the file still
  regenerates byte-for-byte.

- **`run-script` workflow** — drops the clone deploy from the Safe-artifact
  pipeline. It's a deploy-key broadcast (the key deploys the clone, assigns
  the Safe's role grants, and self-revokes admin), not a Safe-signed artifact,
  so it belongs in the manual deploy-key flow.

- **Runbook** (`docs/ETHEREUM_BOOTSTRAP.md`) — operator steps for the Ethereum
  bootstrap.

## Summary by CodeRabbit

* **New Features**
[...]

PR #231, opened 2026-07-07
feat(script): deploy the Ethereum production tokens via the deploy key, matched to Base

hardyjosh:
Stands up the full ST0x production token set on Ethereum in a single
deploy-key broadcast, matched to Base — no Safe signature required.

- **`20260706-deploy-tokens-ethereum.s.sol`** — one operation (`run()`),
  broadcast from the CI deploy key. Mirrors the authoriser-clone deploy
  pattern: the deploy key deploys, configures, then self-relinquishes. Per
  token, in order:
  1. `StoxUnifiedDeployer.newTokenAndWrapperVault` with `initialAdmin` = the
     deploy key and the Base-verbatim name/symbol — the deploy key transiently
     owns the receipt vault (the wrapped vault has no owner).
  2. `setAuthorizer(clone)` on the receipt vault while the key is still owner.
  3. `transferOwnership(Safe)` — single-step OZ `Ownable`, so ownership lands
     on the Safe with no accept step. The deploy key then holds nothing.

  Pre-flight requires the core deployers, the V4 authoriser clone (the
  `setAuthorizer` target) and the Safe (the handoff target) all live, so the
  bootstrap order (Safe → clone → tokens) is enforced. Deployed addresses are
  nonce-based (not known ahead), so it logs each `(underlying, receiptVault,
  wrapped)` for the post-execution pin PR to hydrate
  `productionTokensEthereum()`.

- Name/symbol come from the canonical `LibProdTokenConfig` (introduced in the
  branch below), validated against live Base — SGOV's leading-space name and
  all — so the tokens reproduce Base byte-identically.

- **Workflow** — listed in `manual-broadcast.yaml` (the deploy-key broadcast
  dispatcher), targeting the chosen `network`. It is NOT a Safe-artifact
  script, so it is not in `run-script.yaml`.

Pre-flight guard test fires while the pins are pending; red until the Ethereum
Safe + clone are live, by design.

https://claude.ai/code/session_01VPs1hCTxusmaSeFKvoc4Kr

### DIFF
--- a/.github/workflows/manual-broadcast.yaml
+++ b/.github/workflows/manual-broadcast.yaml
@@ -19,6 +19,7 @@ on:
           # lives in the script's file-level NatSpec — this dropdown is a
           # registry of *which* scripts exist, not *whether* they've run.
           - 20260619-deploy-v4-authoriser-clone
+          - 20260706-deploy-tokens-ethereum
       network:
         description: 'Network to broadcast against (default: base)'
         required: true
--- a/script/20260706-deploy-tokens-ethereum.s.sol
+++ b/script/20260706-deploy-tokens-ethereum.s.sol
@@ -10,10 +10,9 @@ import {
 } from "rain-vats-0.1.6/src/concrete/deploy/OffchainAssetReceiptVaultBeaconSetDeployer.sol";
 import {ReceiptVaultConfigV2} from "rain-vats-0.1.6/src/abstract/ReceiptVault.sol";
 
-import {IGnosisSafe} from "../src/interface/IGnosisSafe.sol";
+import {LibBeaconInvariants} from "../src/lib/LibBeaconInvariants.sol";
 import {IStoxUnifiedDeployerV1} from "../src/interface/IStoxUnifiedDeployerV1.sol";
 import {LibSafeInvariants} from "../src/lib/LibSafeInvariants.sol";
-import {LibProdDeployCurrent} from "../src/generated/LibProdDeployCurrent.sol";
 import {LibProdDeployV4} from "../src/generated/LibProdDeployV4.sol";
 import {LibProdTokenConfig, TokenConfig} from "../src/lib/LibProdTokenConfig.sol";
 
@@ -41,12 +40,12 @@ error DeployerNotDeployed(address deployer);
 /// @param clone The clone address inspected.
 error EthereumCloneNotReady(address clone);
 
-/// @notice Pre-flight failed: the Ethereum token-owner Safe address is not
-/// yet pinned (`address(0)`). The Safe is a distinct per-chain address,
-/// deployed out-of-band and pinned in
-/// `LibSafeInvariants.STOX_TOKEN_OWNER_SAFE_ETHEREUM`; ownership cannot be
-/// handed to it until that address lands.
-/// @param safe The Safe address inspected (`address(0)` when unpinned).
+/// @notice Pre-flight failed: the active chain's resolved token-owner Safe is
+/// not the pinned ETHEREUM Safe. Either the script was dispatched against a
+/// network other than Ethereum mainnet (this script deploys the Ethereum
+/// token set only), or the Ethereum pin
+/// (`LibSafeInvariants.STOX_TOKEN_OWNER_SAFE_ETHEREUM`) has not landed.
+/// @param safe The Safe address the active chain resolved to.
 error EthereumSafeNotReady(address safe);
 
 /// @title DeployTokensEthereum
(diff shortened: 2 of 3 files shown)


hardyjosh:
* **#231** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/231?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#257](https://github.com/S01-Issuer/st0x.deploy/pull/257) <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/257?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#259](https://github.com/S01-Issuer/st0x.deploy/pull/259) <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/259?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/231?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#230** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/230?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#258** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#229** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/229?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#265** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/265?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#266** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/266?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#267** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/267?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#268** <a href="https://app.graphite.com/github/pr/S01-Issuer/st0x.deploy/268?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

thedavidmeister:
Reviewed 12bf0dc30f01714f3447ce5213e9752fd19c0353: approve

Run **29981094045** leaves one test failure — `testDeployTag` (`0_1_8 != 0_1_22`) — plus `git-clean`. Both inherited main-red.

The wiring is right, and the pre-flight is the strongest in the stack: the chain guard reverts unless the resolved Safe *is* Ethereum's (so a misdispatched network cannot deploy a duplicate set elsewhere), the authoriser is checked by codehash rather than a bare `code.length`, in-use beacon ownership is asserted before anything deploys, and the ordering is correct — `setAuthorizer` while the deploy key still owns, then single-step `transferOwnership` so nothing is left pending.

Two gaps found in review, both now fixed on this head:

**No run-once guard.** The sibling clone script refuses to re-run via `V4AuthoriserClonePinAlreadyHydrated`; this had no equivalent. Once the deploy succeeds every other pre-flight passes — deployers live, authoriser live, Safe live — so a re-dispatch would mint **28 more production tokens**, nonce-addressed and indistinguishable from the real ones except by deploy order, and hand them to the Safe. The only thing in the way was a human reading a dropdown that says outright it lists which scripts exist, not which have run. This compounds with the loop shape: `forge script --broadcast` sends one transaction per call, so a failure at token 15 leaves 14 live with no resume, and the only recovery duplicates them. The token table is the record of what a previous run produced, so the guard reads it and refuses on any hydrated leg.

**No post-handoff verification.** `setAuthorizer` and `transferOwnership` are now read back per token. Both happen in the same broadcast, and the acceptance test that would catch a silent miss does not run until the pin PR lands — until then a missed handoff is a production vault owned by a CI key.

Also added `DeploymentEventMissing`, so an unresolved `Deployment` event fails by name mid-broadcast instead of faulting on a zero address.

Both checks take their subject as an argument rather than reaching for pinned state, which is what makes them testable — a check reachable only from inside a broadcast loop can be shown not to fire and never shown to fire. 8 tests, and all three mutants killed against a committed baseline: guard-checks-row-0-only, `receiptVault`-check-disabled, and handoff-assert-disabled (which survived until the assertion was made drivable).

Two issues filed from this review, neither blocking: [rainix#278](https://github.com/rainlanguage/rainix/issues/278) on whether the one-contract-per-file gate should count interfaces, and [rain.vats#330](https://github.com/rainlanguage/rain.vats/issues/330) — `IAuthorizableV1` publishes `authorizer()` but not `setAuthorizer`, which is why this script and the 20260623 one each hand-mirror the same selector under different names.

PR #240, opened 2026-07-08
ci: pass soldeer autopublish secrets explicitly (fix cross-org startup_failure)

thedavidmeister:
Follow-up to #239. The first **Package Release** run (on the #239 merge, commit `1097885d0`) failed with **`startup_failure`** — the run never started.

## Cause

`secrets: inherit` is **not permitted when the reusable workflow is owned by a different organization**. `rainix-autopublish.yaml` lives in `rainlanguage/rainix`; this repo is `S01-Issuer/st0x.deploy` — a different org. raindex can use `inherit` only because it's *in* the rainlanguage org.

This repo's other cross-org rainix callers already pass secrets explicitly — e.g. `rainix-sol.yaml` passes `RPC_URL_*_FORK` by name — and the old (working) `publish-soldeer.yaml` passed `SOLDEER_API_TOKEN` explicitly too.

## Fix

Replace `secrets: inherit` with an explicit block:

- `SOLDEER_API_TOKEN` — required for the publish (already present in this repo; the old workflow used it).
- `PUBLISH_PRIVATE_KEY`, `CI_GIT_EMAIL`, `CI_GIT_USER` — optional; empty falls back cleanly (bot git identity; `GITHUB_TOKEN` push, which works because `main` is **not** branch-protected).
- `CACHIX_AUTH_TOKEN` is intentionally omitted — it's referenced but *not declared* in the reusable's `secrets:` block, so it can't be passed explicitly; nix runs fine without it (as `rainix-sol` already does here).

## On merge

The autopublish runs for real: content-gate current `main` vs published `st0x-deploy@0.1.1` → change → auto-publish **`st0x-deploy@0.1.2`** + tag `sol-v0.1.2` + GH release. I'll watch this run and diagnose (not fire-and-forget).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #239, opened 2026-07-08
ci: soldeer autopublish via rainix-autopublish (unbreak publishing)

thedavidmeister:
## Why

The existing `publish-soldeer.yaml` triggers on `v*` tag push and calls `rainlanguage/rainix/.github/workflows/publish-soldeer.yaml@main` — **that reusable no longer exists in rainix** (only `rainix-autopublish.yaml` remains), so the workflow is dead. The `st0x-deploy` soldeer package is stuck at **0.1.1** (published 2026-05-20); the **0.1.2** orchestrator release (#222) and **0.1.3** corp-actions release (#224) advanced the internal `DEPLOY_TAG` but never reached soldeer, so downstream consumers can't pin the new deploy constants / frozen bytecode.

## How

Adopt the merge-driven autopublish model raindex uses (`rainix-autopublish.yaml`):

- **`.github/workflows/package-release.yaml`** — on push to `main`, calls `rainix-autopublish.yaml@main` with `soldeer-package: st0x-deploy`, `secrets: inherit`. The reusable content-gates the package (dry-run zip vs the latest published revision), and on a real change auto-bumps `foundry.toml [package].version`, publishes to soldeer, tags `sol-v<version>`, and cuts a GH release — **no manual tag**.
- **`foundry.toml`** — add a top-level `[package]` (`name = "st0x-deploy"`, `version = "0.1.1"` = last published) that the content-gate/auto-bump reads.
- **Remove** the dead tag-triggered `publish-soldeer.yaml`.

## Effect on merge

The content-gate compares current `main` against published `st0x-deploy@0.1.1`, sees the 0.1.2 + 0.1.3 changes, and auto-publishes **`st0x-deploy@0.1.2`** (patch bump), catching soldeer up with the on-chain deploy constants. Subsequent merges publish only on real content change.

Soldeer semver (`0.1.x`) is intentionally decoupled from the internal `DEPLOY_TAG` (`0_1_3`) — same as raindex's soldeer `0.1.13` vs its own deploy tags. st0x's deploy-constant pins are not self-referentially rewritten on publish, so there is no bump loop.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added an automated release flow for package publishing on updates to the main branch.
  * Updated the package identity and version information to reflect the latest release.
* **Chores**
  * Removed the older tag-based publishing workflow.

PR #241, opened 2026-07-08
ci: grant Package Release the write permissions autopublish needs

thedavidmeister:
Follow-up to #239 / #240. Both Package Release runs so far failed with **`startup_failure` and zero jobs**, and #240's `secrets` fix didn't help because **that wasn't the cause**.

## Real root cause

`rainix-autopublish`'s `release` job declares `permissions: { contents: write, id-token: write }` (it pushes the version-bump commit + tag and cuts a GH release). This repo's **default workflow-token permission is read-only**:

```
$ gh api repos/S01-Issuer/st0x.deploy/actions/permissions/workflow
{ "default_workflow_permissions": "read", ... }        # raindex = "write"
```

When the caller's token is read-only, a reusable requesting `write` **exceeds** what the caller can grant, so the run fails at startup before any job. raindex succeeds with the identical reusable only because its repo default is `write`.

## Fix

Grant exactly the scopes the reusable needs on the caller job:

```yaml
jobs:
  release:
    permissions:
      contents: write
      id-token: write
    uses: rainlanguage/rainix/.github/workflows/rainix-autopublish.yaml@main
```

This keeps the repo's **read-only default intact** (least privilege) rather than flipping the whole repo to `write` like raindex. `allowed_actions` is already `all`, so no other policy blocks it.

## On merge

The autopublish should finally run for real: content-gate current `main` vs published `st0x-deploy@0.1.1` → change → publish **`st0x-deploy@0.1.2`** + tag `sol-v0.1.2` + GH release. I'll watch this run and diagnose.

### DIFF
--- a/.github/workflows/package-release.yaml
+++ b/.github/workflows/package-release.yaml
@@ -5,15 +5,22 @@ on:
       - main
 jobs:
   release:
+    # rainix-autopublish's job requests contents: write (to push the version-
+    # bump commit + tag) and id-token: write. This repo's default workflow-token
+    # permission is read-only, so without granting these here the reusable's
+    # request exceeds the caller and the run fails at startup. Grant exactly the
+    # scopes the reusable needs, per workflow, rather than flipping the whole
+    # repo default to write.
+    permissions:
+      contents: write
+      id-token: write
     uses: rainlanguage/rainix/.github/workflows/rainix-autopublish.yaml@main
     with:
       soldeer-package: st0x-deploy
-    # Pass secrets EXPLICITLY, not `inherit`: GitHub disallows `secrets: inherit`
-    # when the reusable is owned by a different org (rainlanguage vs S01-Issuer),
-    # which fails the run at startup. Mirror this repo's other cross-org rainix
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 085aad1c7 — **Claude Code agent review** (automated review artifact, not a human approval; the merge is authorized by @thedavidmeister, who instructed it).

Single-file CI-config fix: adds `permissions: { contents: write, id-token: write }` to the Package Release caller job. Root-caused the two prior `startup_failure` runs to this repo's read-only default workflow-token permission (`gh api .../actions/permissions/workflow` = `read`; raindex = `write`) vs `rainix-autopublish`'s write request. Grants exactly the scopes the reusable needs, per workflow, preserving the repo's read-only default (least privilege). `allowed_actions` is already `all`. git-clean/static/legal/test/CodeRabbit green.

Residual risk: an org-level token-permission hard cap (not visible from here) would still block explicit elevation; if this run startup_fails again, the fallback is flipping the repo default to write.

Agent-review outcome: no blocking issues at 085aad1c7.

PR #246, opened 2026-07-10
fix(invariants): reject null vault authoriser unconditionally in migration invariant

### LINKED ISSUE #245: Fail-open: production invariant bundle accepts a null (address(0)) vault authoriser
## Summary
`LibTokenInvariants.assertUniformAuthoriserMigration` — and via it `LibInvariants.assertAll` (the production invariant bundle the cron enforces) — **fails open on a null authoriser**. A production receipt vault whose `authorizer()` is `address(0)` (a bricked / ungated vault — exactly what this invariant exists to catch) **passes** today.

## Why
The migration-window check calls `LibMigrationInvariant.assertMigration(actual = vault.authorizer(), pre = STOX_PROD_AUTHORISER (V3), post = STOX_PROD_AUTHORISER_V4_CLONE, deadline)`. The V4 clone pin is still `address(0)` (not yet hydrated), so `post == address(0)`. Before the deadline the guard reverts only when `actual != pre && actual != post`. For a vault with `authorizer() == address(0)`:

- `actual != pre` → `0 != V3` → true
- `actual != post` → `0 != address(0)` → **false**
- `true && false` → **false → no revert → passes**

i.e. a null authoriser collides with the still-zero `post` value and is accepted as 'already migrated to V4'.

## Impact
**MEDIUM.** This bundle is the cron guard meant to catch a misconfigured/bricked vault; today a vault with no authoriser reads green. Live now across `assertAll`. Operator-facing invariant (not directly exploitable), but it defeats the guard's purpose.

## Proposed fix
In `assertUniformAuthoriserMigration` (or the right layer), reject `actual == address(0)` **unconditionally**, independent of the migration window, with a specific typed error. Preserve legitimate window behavior: a real V3 `pre` and a real V4 `post` both pass before the deadline; only V4 after.

## Related gaps in the same area
- **`productionReceiptVaults()` has no exhaustiveness guard** — a vault added to the constants but omitted from the hardcoded ~20-entry array is silently skipped by the swap bundle AND every token invariant, and cron stays green.
- **Zero-sentinel path untested** — every existing test uses non-zero sentinels (0xAAAA/0xBBBB), which is exactly why this fail-open went uncaught. A test with `pre`/`post`/`actual == address(0)` is the regression guard.

## Status
A mutation-validated fix + tests are prepared and can land as a PR closing this. Surfaced by the ramifications audit of #197 (whose migration pins call this bundle).

_File/line refs approximate (`LibTokenInvariants.sol` ~293-300, `LibInvariants.sol` ~52-56, `LibMigrationInvariant.assertMigration`)._

thedavidmeister:
## What

`LibTokenInvariants.assertUniformAuthoriserMigration` — and via it `LibInvariants.assertAll`, the production invariant bundle the cron enforces — failed open on a null (`address(0)`) vault authoriser. The V4 post pin (`LibProdDeployV4.STOX_PROD_AUTHORISER_V4_CLONE`) is still unhydrated `address(0)`, so for a vault whose `authorizer()` is `address(0)` the window guard's boolean derivation reads:

- `actual != pre` → `0 != V3` → true
- `actual != post` → `0 != 0` → **false**
- `true && false` → false → **no revert → a bricked, ungated vault passes as "already migrated"**

This PR rejects `actual == address(0)` **unconditionally** — before the migration window is consulted and independent of the deadline — with a new typed error `ReceiptVaultNullAuthoriser(address vault)` that surfaces the offending vault, declared alongside the sibling `ReceiptVault*` invariant errors.

## Layering

The rejection lives in `assertUniformAuthoriserMigration` (the semantic authoriser leg), **not** in the generic `LibMigrationInvariant.assertMigration`: that helper is a reusable dual-state window primitive with `bytes32`/`uint256` overloads serving migrations where zero can be a legitimate value on either side. Only the authoriser leg knows that a production vault is never legitimately ungated, so that is where "null is never acceptable" belongs. Legitimate window behaviour is untouched: real `pre` and real `post` both pass before the deadline, only `post` at/after it.

Touched files are the operator-tooling invariant layer (`src/lib/LibTokenInvariants.sol`) plus tests — no deployed vault source, no generated deploy libs, no bytecode change.

## QA

- Discriminating tests: `testNullAuthoriserRevertsBeforeDeadline`, `testNullAuthoriserRevertsAfterDeadline` (in new `test/src/lib/LibTokenInvariantsMigration.t.sol`, using the CURRENT production pins: pre = real V3 authoriser, post = `STOX_PROD_AUTHORISER_V4_CLONE` = `address(0)`, deadline = `V4_SWAP_DEADLINE`) — each fails on base: with the zero check removed (base behaviour), both fail with "next call did not revert as expected", i.e. the fail-open passes-without-revert was reproduced and proven. Window-preservation tests `testWindowAcceptsPreBeforeDeadline` / `testWindowAcceptsPostBeforeDeadline` / `testWindowAcceptsPostAfterDeadline` / `testWindowRejectsPreAfterDeadline` (non-zero sentinels) pass on base and on the fix by design — they pin that the null rejection does not narrow legitimate window behaviour and kill over-correction mutants.
- Mutations applied:
  - remove the zero check → `testNullAuthoriserRevertsBeforeDeadline` + `testNullAuthoriserRevertsAfterDeadline` fail (killed)
  - window-gate the check to `block.timestamp < deadline` → `testNullAuthoriserRevertsAfterDeadline` fails (killed)
  - reverse-gate to `block.timestamp >= deadline` → `testNullAuthoriserRevertsBeforeDeadline` fails (killed)
  - surface `address(0)` instead of the offending vault → both null tests fail on exact revert-data mismatch (killed)
- Oracle: the issue's boolean derivation of the window guard (`actual != pre && actual != post` with `post == 0` collapsing to accept `actual == 0`) plus migration semantics — a null authoriser is never a legitimate state on either side of the swap — derived independently of the implementation.
- Category check: issue asks unconditional null-authoriser rejection with a specific typed error + zero-sentinel regression tests; both covered (zero-sentinel path exercised before AND after the deadline under the exact production pin configuration). The `productionReceiptVaults()` exhaustiveness guard is a related gap the issue lists, not the ask — explicitly out of scope here (separate engineering change).

Closes #245

Co-Authored-By: Claude <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a dedicated revert error when a production receipt vault reports a null authorizer.
  * Null authorizers are now rejected unconditionally, with behavior consistent both before and after the migration deadline.
  * Continued to enforce migration window/deadline correctness for invalid authorizer transitions.

* **Tests**
  * Added migration invariant tests covering null authorizer reverts before/after the deadline.
  * Added tests for accepted valid pre/post authorizer states and rejection when a “pre” vault is detected after the deadline.


thedavidmeister:
🤖 ai:vetter
Reviewed 4e2ba7ea116786cd50161c06c7d0fcbcbf28eedc: ready — closes #245 — fail-open verified against source (post pin STOX_PROD_AUTHORISER_V4_CLONE is address(0), assertMigration accepts actual==post before deadline), unconditional ReceiptVaultNullAuthoriser guard at the correct layer (authoriser leg, generic bytes32/uint256 window helper untouched), both null tests discriminate under real prod pins, window behavior pinned by 4 sentinel tests, mutation claims all check out; note issue's related productionReceiptVaults() exhaustiveness gap stays open as separate work
cost 478 — invariant fail-open fix, prod-pin and window semantics review

thedavidmeister:
**Design decision: keep the caller-level null-authoriser check (this PR), not a zero-`post` rejection in the generic `assertMigration`.**

Considered moving the guard into `LibMigrationInvariant.assertMigration` by rejecting a zero `post`. It would fix the fail-open at the source and would NOT affect the codehash-migration layer (its `post` is a real pinned codehash; only its `pre` is `bytes32(0)`, for an undeployed impl). But rejecting zero `post` was declined because `post == address(0)` is a **deliberate live state** of the authoriser leg:

- `LibProdDeployV4.STOX_PROD_AUTHORISER_V4_CLONE` is unhydrated (`address(0)`), and `LibInvariants.assertAll` composes the authoriser migration with that zero `post` on purpose.
- Before `V4_SWAP_DEADLINE` the window passes on the V3 `pre` side — the invariant is cron-covered from merge, before the clone is even deployed.
- After the deadline the unsatisfiable `must-be-post` (== 0) branch is the **forcing function** that red-lines an un-run swap.

Rejecting a zero `post` would force the authoriser leg to be gated on hydration (`if (post != 0)`) and would lose that deadline forcing-function. So the null-authoriser rejection stays where the domain knowledge lives — in `assertUniformAuthoriserMigration`, before it delegates — with its own typed `ReceiptVaultNullAuthoriser` error. `assertMigration` stays domain-agnostic (zero is a legitimate `pre` in the codehash layer). Closing the design question; keeping this fix as-is.

thedavidmeister:
Closing: the defect this PR fixes no longer exists.

The premise was that `STOX_PROD_AUTHORISER_V4_CLONE` was unhydrated `address(0)`, so the migration window guard failed open — `0 != V3` true, `0 != 0` false, and a bricked ungated vault passed as "already migrated". On main that pin is now hydrated:

```
src/generated/LibProdDeployV4.sol:141
address constant STOX_PROD_AUTHORISER_V4_CLONE = address(0x315b16faa6eE413faBCa877d3851B3818369f0cD);
```

With a non-zero post pin, `0 != 0x315b…` is true and a null authoriser already reverts. The bug was real when this was opened on 2026-07-10; hydration closed it.

Worth recording why the green CI here was not evidence the change still worked as described: `testNullAuthoriserRevertsBeforeDeadline` / `testNullAuthoriserRevertsAfterDeadline` were built on `post = STOX_PROD_AUTHORISER_V4_CLONE = address(0)`. That equality no longer holds, so base reverts too and the tests stopped discriminating. They pass — they just no longer prove anything about base. Green meant "passes", not "would fail without the fix".

The residual concern is not the null check itself but that it is now only **incidentally** true: it holds because the pin happens to be non-zero, and this repo sat in the fail-open state for weeks. A future migration that introduces an unhydrated post pin reinstates the same hole. That is a pin-drift hazard rather than a live bug, and it belongs on #245 rather than in a PR whose justification has expired.

Not deleting the branch.

PR #261, opened 2026-07-16
Move the Protofire audit under audit/protofire/ so the scan finds it

thedavidmeister:
st0x.deploy has a Protofire audit, but the org-health dashboard reads it as **never externally audited** — the report was committed at `audit/Report_st0x.deploy_may_2026_4.0.pdf`, and roh-scan only counts PDFs under **`audit/protofire/`**.

This `git mv`s it to `audit/protofire/st0x.deploy.v0.1.1.may-2026.pdf`:
- **Under the scanned path** so it's counted.
- **Named to the scan's convention** (`<repo>.<anchor>.<month>-<year>`), tag-anchored to **`v0.1.1`** so drift is measured `compare/v0.1.1...HEAD`. Verified roh's regex extracts `v0.1.1`.

Pure rename (100% similarity, content unchanged). REUSE stays satisfied (`REUSE.toml` covers `audit/**/`).

After it lands, the next scan moves st0x.deploy off `never` to audited-at-`v0.1.1` with accrued drift.


thedavidmeister:
Reviewed 1a440996b38dac3e573b06d0de0a240d044d46b3: approve

Read the diff myself: a single 100%-similarity rename of the Protofire report from audit/ to audit/protofire/st0x.deploy.v0.1.1.may-2026.pdf, tag-anchored to v0.1.1 so roh-scan counts it. Zero content change; touches nothing but that one file.

Merging over two reds, both confirmed PRE-EXISTING on main and both orthogonal to a PDF move:
- git-clean — main itself fails it (stale src/generated pointer artifacts; a 0_1_10 release's pointers aren't regenerated on main). This PR does not touch src/generated.
- rainix-sol/test — main's latest run (8cae29b5c) is already failure. A doc-only rename cannot change compilation or fork tests.

legal, static, CodeRabbit pass; no unresolved threads.

PR #259, opened 2026-07-16
pin(tokens): hydrate the Ethereum token table + add RKLB to the canonical set

hardyjosh:
Pins the full Ethereum production token table and completes the canonical set with RKLB (previously an accidental omission from every table in this repo). PINS ONLY — the RKLB operational scripts live in the follow-up ops PR (#271).

**Ethereum table hydration.** The `20260706-deploy-tokens-ethereum` broadcast executed 2026-07-22 (run 29921218929): 28 tokens deployed via the 0.1.1 unified deployer, each wired onto the Ethereum V4 authoriser and handed to the Ethereum token-owner Safe — 84 txs, zero failures. `productionTokensEthereum()` is hydrated from the run's logged `(underlying, receipt, receiptVault, wrapped)` tuples; first + last vaults spot-verified live.

**RKLB pins.** Live production on Base (Safe-deployed at block 41704399, Safe-owned, V1-address production beacons) but absent from every table — so no invariant covered it, the V4 swap batch skipped it (still on the V3 authoriser), and the Ethereum broadcast never deployed it. This PR adds the `LibProdTokenConfig` row + Base table row 29 (validated against the live Base vault), and the Ethereum row 29 pinned from the executed gap-fill broadcast (run 29924926246), verified wired live.

The full 29-token Ethereum parity leg passes against live mainnet. The strict Base uniform-authoriser invariants stay red until the two pending Safe bundles (6-vault + RKLB single-tx, authored via #271's scripts) execute — the intended forcing function.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VPs1hCTxusmaSeFKvoc4Kr

## Summary by CodeRabbit

* **New Features**
  * Added Rocket Lab USA Inc. (RKLB) token support across production environments.
  * Added the RKLB token with symbol **tRKLB** to cross-chain token configurations.
  * Updated production token listings and deployment records to include the new token.

PR #279, opened 2026-07-23
Test the run-once guard against a table the test controls

thedavidmeister:
main is red, and it is my fault.

The run-once guard added in [#231](https://github.com/S01-Issuer/st0x.deploy/pull/231) is fine. Its tests were not: they read `productionTokensEthereum()` and hydrated a row of the **real** table. That asserted a fact about the afternoon they were written — the shipped table was all placeholders — rather than a property of the guard.

[#259](https://github.com/S01-Issuer/st0x.deploy/pull/259) hydrated the table, and all four inverted at once:

```
testRunRevertsWhenCoreNotDeployed       expected DeployerNotDeployed,
                                        got EthereumTokensAlreadyDeployed (the gate now fires first)
testRunOnceGateFiresOnAHydratedTable    expected 0xBEEF at row 7, got row 0`s real address
testRunOnceGateInspectsEveryLeg         expected 0xC0DE at row 27, same
testRunOnceGatePassesOnTheShippedTable  asserted the shipped table is virgin — it is not
```

`main` has been red since `b2e3fa0`.

## Fix

The guard tests build their own three-row table via a `virginTable(...)` helper, so what the production table holds can no longer reach them. Same assertions, same coverage, no dependency on a moving fact.

`testRunRevertsWhenCoreNotDeployed` is gone — the run-once gate legitimately fires before the deployer check now, so that ordering can never be observed again. In its place, `testShippedTableIsHydratedSoTheScriptIsSpent` asserts what is true today: the shipped table is hydrated, the guard refuses it, and `run()` cannot proceed on any chain. The script is permanently self-disarmed, which is the property actually worth pinning — and unlike the assertion it replaces, it cannot go stale, because hydration is one-way.

8 tests pass, `forge fmt --check` clean.

## The lesson, since it is the same one this stack kept surfacing

A test that reads live or shipped state and asserts today`s value is a snapshot wearing a test`s clothes. It is the same defect as a `PENDING` marker after execution, a happy path that authors against an empty target set, and a live-fork verdict quoted hours later — all of which turned up in this stack. I fixed those and then wrote this. The guard was right; the way I checked it was not.

### DIFF
--- a/test/script/20260706-deploy-tokens-ethereum.t.sol
+++ b/test/script/20260706-deploy-tokens-ethereum.t.sol
@@ -29,17 +29,31 @@ import {LibStoxDeployNetworks} from "../../src/lib/LibStoxDeployNetworks.sol";
 /// 0.1.1 deployers, in-use beacon ownership, then the clone (setAuthorizer
 /// target), then the Safe (handoff target).
 contract DeployTokensEthereumTest is Test {
-    /// `run()` reverts `DeployerNotDeployed` when the 0.1.1 core has not been
-    /// broadcast to the active chain — the unified deployer's pinned address
-    /// has no code, so no token can be deployed. (No fork: the deployer is
-    /// absent, which is the pre-bootstrap state, and is the first guard in
-    /// the pre-flight chain.)
-    function testRunRevertsWhenCoreNotDeployed() external {
-        DeployTokensEthereum script = new DeployTokensEthereum();
-        vm.expectRevert(
-            abi.encodeWithSelector(DeployerNotDeployed.selector, LibProdDeployV4.STOX_UNIFIED_DEPLOYER_0_1_1)
-        );
-        script.run();
+    /// A virgin table, built here rather than read from the shipped one. The
+    /// guard's behaviour is a property of the guard, not of whatever the
+    /// production table happens to hold today — reading the real table made
+    /// these tests assert a moment in time, and they broke the moment the
+    /// Ethereum pin PR hydrated it.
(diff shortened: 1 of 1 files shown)


thedavidmeister:
Reviewed 1b0a81e: approve

Fixes main-red that I caused in #231.

The run-once guard is correct. Its tests were not: they read `productionTokensEthereum()` and hydrated a row of the **real** table, asserting a fact about the afternoon they were written rather than a property of the guard. #259 hydrated that table and all four inverted at once — the gate now trips on row 0 before reaching the staged row, and one test asserted the shipped table was virgin, which stopped being true.

They now build their own three-row table, so production state cannot reach them. `testRunRevertsWhenCoreNotDeployed` is replaced by `testShippedTableIsHydratedSoTheScriptIsSpent`, which pins what is true now: the guard refuses the shipped table and `run()` cannot proceed on any chain. That assertion cannot go stale the way its predecessor did, because hydration is one-way.

8 tests pass locally, `forge fmt --check` clean.

**Merging before CI reported.** Stating that plainly rather than implying otherwise: the change is test-file-only, verified locally, and `main` is currently red for every PR in the repo — this cannot make that worse, and leaving it red has a running cost. If CI on `main` disagrees I will fix forward.

The lesson is the one this stack kept surfacing and I then repeated: a test that reads live or shipped state and asserts today's value is a snapshot in a test's clothing — the same defect as a `PENDING` marker after execution, a happy path authoring against an emptied target set, or a live-fork verdict quoted hours later. I corrected three of those in this stack and wrote a fourth.

PR #300, opened 2026-08-11
chore(deps): bump OpenZeppelin 5.6.1 → 5.7.0 — only the wrapped vault's bytecode moves

thedavidmeister:
## What

Bumps `@openzeppelin-contracts` and `@openzeppelin-contracts-upgradeable` soldeer deps 5.6.1 → 5.7.0, sweeps the versioned import prefixes across src/test/script, regenerates the candidate pointers to a fixpoint, and records the result in the CHANGELOG.

## The byte-level answer this PR exists to give

- **Governance timelock: byte-identical.** `testTimelockPinsMatchCompiledDependency` passes unchanged against the 5.7.0 dependency — the frozen `TIMELOCK_CREATION_CODE` / `TIMELOCK_RUNTIME_CODEHASH` pins (and therefore the three live deployments) are reproduced exactly. Independently cross-checked by compiling both tags standalone under the repo's exact profile: identical 8,149-byte creation code, identical runtime keccak `0xb623…6993`.
- **Exactly one contract's bytecode moves: `StoxWrappedTokenVault`.** 5.7.0's `ERC4626Upgradeable` swaps its private `_tryGetAssetDecimals` for the shared `SafeERC20.tryGetDecimals` helper. The candidate pointer regeneration cascades through everything embedding the vault's Zoltu address: its beacon, the beacon-set deployer, the unified deployer, and the orchestrator pair. Frozen release-tag snapshots are untouched — nothing deployed changes until a release is cut, at which point this is a new-tag event per the versioning rules.
- Everything else (receipt vault, receipt, authoriser machinery, deployers on the receipt side) compiles byte-identical.

## Reviewer decision

rain-vats 0.1.6 still imports `@openzeppelin-contracts-5.6.1/` prefixes internally, so two `remappings.txt` lines bridge those onto the 5.7.0 install (same class of bridge as the existing `@openzeppelin/contracts/` line, documented in the CHANGELOG; soldeer regenerating the file would need them re-applied). The alternative is waiting for rain-vats to republish against 5.7.0 and bumping in lockstep — this PR takes the bridge; call it out if you'd rather wait.

## QA

- Discriminating tests: `testCandidateSelfConsistent` — fails on base source combined with this branch's regenerated pointers, and was observed failing live on this branch before the pointer fixpoint (candidate constants vs 5.7.0 compile mismatch); `testTimelockPinsMatchCompiledDependency` deliberately does NOT discriminate — its passing under 5.7.0 IS the byte-identity claim.
- Mutations applied: reverting any regenerated pointer constant to its 5.6.1 value → killed by `testCandidateSelfConsistent` (observed directly: the pre-fixpoint tree, where the beacon pointer still carried the 5.6.1-era embedded vault address, failed exactly there with the mismatched bytecode in the assertion output). Dep-version line itself: reverting foundry.toml to 5.6.1 alone → build breaks on the swept import prefixes (compile-time kill).
- Oracle: the compiled artifacts of the version-locked OZ deps under the repo's single `[profile.default]` (solc 0.8.25, optimizer 2000, cancun, no metadata) — pointer expectations come from the compiler over the dependency, never hand-written; for the timelock additionally the three live on-chain deployments the frozen pins describe.
- Category check: no linked issue — scope is the bump itself: both OZ deps moved together, import sweep verified grep-zero on old prefixes (sole survivor is the frozen-pin provenance doc, deliberate), pointer cascade regenerated to fixpoint (3 passes), CHANGELOG entry added.

## Summary by CodeRabbit

* **Updates**
  * Upgraded OpenZeppelin Contracts and Upgradeable Contracts to version 5.7.0.
  * Documented resulting bytecode and deployment address changes, while preserving timelock configuration pins.
  * Added compatibility guidance for remaining legacy imports.

* **Documentation**
  * Clarified deployment table handling: extra canonical rows are supported, while shorter or misaligned tables now revert with dedicated errors.


thedavidmeister:
🤖 ai:producer
Design-question: Static is still red at ac6d5b6 and I could not establish why, so I am routing the one thing here that needs a human ruling rather than guessing.

WHAT IS FIXED (verified)
- The DIRTY conflict is resolved: origin/main (75 commits) merged in. Sole conflict was src/lib/LibBeaconInvariants.sol - this branch bumped the OZ import prefix, main added a LibMigrationInvariant import; both kept.
- git-clean is GREEN. Its cause was that src/generated/LibProdDeployCurrent.sol and src/generated/LibProdDeployV4.sol were committed unformatted. Locally, forge fmt named exactly those two files, and the full CI sequence (soldeer install -> BuildPointers -> fmt -> git diff --exit-code) now exits 0 and is idempotent. BuildPointers regenerated NO pointer content, so the bump needs no pointer changes beyond those already committed.

THE DESIGN QUESTION: which Solidity toolchain is canonical?
This repo runs its two Solidity checks under two different, mutually inconsistent toolchains:
- .github/workflows/git-clean.yaml: `nix develop github:rainlanguage/rainix#sol-shell` - UNPINNED, tracks rainlanguage/rainix main.
- .github/workflows/rainix-sol.yaml (the `static` job): the rainix reusable workflow, which runs `nix develop -c rainix-sol-static` against THIS repo flake. flake.nix inputs github:rainPROTOCOL/rainix, and flake.lock pins rev 8603a225676b3d4d6d7cd0e41ae8918aa2ee4bf5 - a different GitHub owner, and a rev old enough that it provides no sol-shell output at all (nix develop github:rainprotocol/rainix/8603a225...#sol-shell errors with "does not provide attribute devShells.x86_64-linux.sol-shell").

The pipeline own tooling already refuses this: `pr-review-report sol-toolchain` exits 3 on this checkout with "this checkout own push/PR checks run 2 different Solidity toolchains ... A `forge fmt` that satisfies one can fail another, so there is no single answer to give."

I formatted with git-clean toolchain. git-clean went green; static did not. Whether static residual failure is a fmt disagreement between the two forge builds, or an unrelated slither finding, I DID NOT VERIFY - see below. Either way the two-toolchain split is a real defect that makes "format so both checks pass" unanswerable, and choosing between (a) bumping flake.lock to rainlanguage/rainix main or (b) pinning git-clean.yaml to flake.lock rainix rev is a repo-owner call, not mine.

WHY I DID NOT ESTABLISH THE STATIC CAUSE (tooling gaps, not choices)
1. Reading the failing job log needs `gh run view --log-failed`; the permission system denied it, and pr-review-report has NO log-reading subcommand. That is a genuine gap - `await` names WHICH check failed but never WHY.
2. Local reproduction of `nix develop -c rainix-sol-static` stalled indefinitely downloading a multi-GB dev-shell closure (chromium, gtk3, ffmpeg) and never reached slither.

Nothing about the static failure is asserted here beyond the observed fact that it is FAILURE at ac6d5b6 while test, legal, build-artifact and git-clean pass.

PR #308, opened 2026-08-13
codegen: use LibCodeGen.addressConstantString instead of a local copy

### LINKED ISSUE #307: BuildPointers hand-rolls an address-constant emitter LibCodeGen already provides
`script/BuildPointers.sol` hand-rolls an address-constant emitter that `rain-sol-codegen` already publishes.

## The duplication

This repo's build script defines its own local emitter for `address constant DEPLOYED_ADDRESS = address(...)`, while `LibCodeGen` already provides:

```solidity
// rain.sol.codegen src/lib/LibCodeGen.sol:298
function addressConstantString(Vm vm, string memory comment, string memory name, address data)
```

The shared version is strictly more general — parameterized over the comment text and the constant name rather than hardcoded to one name.

The script already depends on `rain-sol-codegen` and already calls `LibCodeGen.bytecodeHashConstantString` for the codehash constant. So it uses the shared library for one generated constant and a private reimplementation for the other, in the same file.

## Scope: four repos, one shared function

Measured 2026-08-13, repos emitting `address constant DEPLOYED_ADDRESS` from a local helper rather than `LibCodeGen`:

| repo | script |
|---|---|
| rainlanguage/rain.deploy | `script/Build.sol` — **being fixed in rain.deploy#26** |
| rainlanguage/rain.factory.deploy | `script/BuildPointers.sol` |
| rainlanguage/rain.verify | `script/BuildPointers.sol` |
| S01-Issuer/st0x.deploy | `script/BuildPointers.sol` |

rain.math.float's build script has no such emitter.

## Ask

- [ ] Replace the local emitter with `LibCodeGen.addressConstantString`, passing the current NatSpec as `comment` and `DEPLOYED_ADDRESS` as `name`
- [ ] Check whether this repo's local copy has DRIFTED from the shared one in comment text, spacing or line wrapping — a drifted emitter changes generated output, and the generated file is committed, so any difference must be regenerated in the same commit or the git-clean gate catches it
- [ ] Report the difference rather than silently changing generated bytes to fit

If `LibCodeGen`'s version genuinely cannot produce the required output, extend `rain-sol-codegen` rather than keeping a private copy — that is a PR there plus a version bump here.

## Why this matters beyond tidiness

The rule is: use the codegen library for codegen, or extend it. Four private copies of one function means four places a generated-output convention can drift apart, in repos whose generated files are frozen deploy pins consumers import.

It is also a symptom of the per-repo build script that rainlanguage/rainix#304 is about — each deploy repo owning its own generator means each one re-solves the same problem privately, and a shared library 300 lines long is easy to miss when a local helper is twenty.

thedavidmeister:
Closes #307

## Motivation

`script/BuildPointers.sol` defined its own `addressConstantString` for
`address constant DEPLOYED_ADDRESS = address(...)` while, in the same file,
calling `LibCodeGen.bytesConstantString` from `rain-sol-codegen` for the other
generated constants. `LibCodeGen` already publishes the address emitter, and its
version is strictly more general — parameterised on the comment text and the
constant name instead of hardcoding both.

## Changes

- `script/BuildPointers.sol`: local `addressConstantString` deleted;
  `buildContractPointers` now uses `LibCodeGen.addressConstantString(vm, <the
  same NatSpec>, "DEPLOYED_ADDRESS", deployed)`.
- `rain-sol-codegen` `0.1.0` -> `0.1.3` (`foundry.toml`, `soldeer.lock`,
  `remappings.txt`, the two versioned import prefixes). Required: the pinned
  `0.1.0` predates `addressConstantString`, which first shipped in published
  `0.1.2`. The now-dangling `rain-sol-codegen-0.1.0/` remapping is dropped;
  `forge soldeer update` adds the new entry but does not remove the stale one.

## Why 0.1.3 and not latest

`0.1.0 -> 0.1.3` is **purely additive** in `src/lib` — it adds
`addressConstantString`, `bytes32ConstantString` and `LibSnapshot.sol` and
changes nothing this repo already calls, so generated output cannot move.

`0.1.4` is not a candidate here, and matters more in this repo than in the other
two. It changes `LibFs.pathForContract` from `<name>.pointers.sol` to
`<name>.sol` and rewrites the `THIS FILE IS AUTOGENERATED BY
./script/BuildPointers.sol` header emitted into every generated file. Taking it
would rename and rewrite files inside the **frozen** `src/generated/0_1_1/`
snapshot, which downstream pins, and would break `LibProdDeployV4.sol`'s
`./<tag>/<name>.pointers.sol` import paths. That is a separate migration, not a
side effect of deduplicating one function.

## Drift check: byte-identical, nothing regenerated

The issue asks whether this repo's private copy had drifted from the shared one.
It had not. Both emit

```
\n<comment>\naddress constant DEPLOYED_ADDRESS = address(0x…);\n
```

and the shared function's only extra behaviour is a line-wrap branch that fires
above 120 columns. This line is 88 columns
(`17 + 16 ("DEPLOYED_ADDRESS") + 3 (" = ") + 8 ("address(") + 42 + 2 (");")`), so
the single-space branch is taken and the bytes match exactly.

Confirmed empirically, not just by reading: `forge script ./script/BuildPointers.sol
&& forge fmt` on this branch leaves `src/generated/` **completely clean** — the
12 `candidate/` pointer files, `LibProdDeployV4.sol` and
`LibProdDeployCurrent.sol` all regenerate to the bytes already committed.

**No frozen snapshot is rewritten.** `deployTag()` is the rolling `candidate`
tag, so the script only ever writes `src/generated/candidate/`; the frozen
numbered snapshot `src/generated/0_1_1/` is read (to emit its aliases into
`LibProdDeployV4.sol`) and never written. It came through the regeneration
untouched, and CI's `frozen-snapshots-append-only` check enforces that
independently.

## QA

- Discriminating tests:
  - The repo's own `Git is clean` workflow, run verbatim locally: `forge script
    ./script/BuildPointers.sol && forge fmt && git diff --exit-code`. It is
    discriminating for exactly this change because the committed
    `src/generated/candidate/*.pointers.sol` were produced by the OLD
    hand-rolled emitter, so any byte the new one emits differently shows up as a
    diff.
  - Baseline run recorded `script_rc=0 generated_files_dirty=0` — proof the gate
    actually executed and was green, not merely assumed.
- Mutations applied:
  - Three mutants on the new call site, one per drift axis the issue names, each
    regenerated and re-checked against the committed artifacts:
  - M1 comment text (`deploy address` -> `deployed address`): KILLED,
    `generated_files_dirty=12`.
  - M2 constant name (`DEPLOYED_ADDRESS` -> `DEPLOYED_ADDR`): KILLED,
    `generated_files_dirty=12`.
  - M3 line wrapping (the comment's internal `\n` collapsed to a space): KILLED,
    `generated_files_dirty=12`.
  - 3/3 killed, 0 survived, each mutant reaching all 12 generated contracts. The
    harness aborts if the unmutated baseline does not come back clean, so a
    silently-not-running gate cannot report kills.
- Oracle:
  - The committed generated artifacts themselves, which are the pre-change
    output of the emitter being replaced. Byte-equality against them is the
    whole correctness claim; it is checked, not reasoned about.
- Category check:
  - The drift axes are covered as a category, not as three examples: comment
    text, constant name and line wrapping are the complete set of inputs the
    shared function's output depends on, since it is `pure` over
    `(comment, name, address)` with one length-derived branch. The address
    itself is covered by the unchanged `BYTECODE_HASH`/`CREATION_CODE`/
    `RUNTIME_CODE` constants generated beside it, and the downstream
    `LibProdDeployV4`/`LibProdDeployCurrent` aliases are regenerated in the same
    run. No production Solidity changes in this PR: the only non-script edits
    are the dependency pin, its lock and the remappings it generates, so no
    deployed bytecode and no deterministic address moves.

## Summary by CodeRabbit

* **Improvements**
  * Updated the code generation tooling to a newer version.
  * Improved generated address constant formatting for build outputs.


thedavidmeister:
🤖 ai:vetter
vet-protocol 4
lens source@7e8c9f997a182acaa8f9f724be407aa85311d322 + audit skill invoked at pr:308
Reviewed 7e8c9f997a182acaa8f9f724be407aa85311d322: ready — Closes #307 — local addressConstantString deleted (zero remaining hits), DEPLOYED_ADDRESS now via LibCodeGen.addressConstantString with a comment literal matching the committed candidate pointers exactly (88-col constant line, under the 120 wrap threshold), so output is byte-identical — structurally proven by the green git-clean CI job that reinstalls deps, regenerates and diffs; rain-sol-codegen 0.1.0→0.1.3 is additive, the dangling 0.1.0 remapping is dropped, nothing under src/generated/ moves, and 0.1.4's file-renaming migration is correctly excluded.
cost 268 — codegen dedupe plus additive dep bump

PR #312, opened 2026-08-15
Remove the submodule-era residue: dangling REUSE.toml entries, three dead .soldeerignore lines, and the CLAUDE.md bump runbook calling soldeer pins git submodules

### LINKED ISSUE #311: Remove the submodule-era residue — dangling `REUSE.toml`/`.soldeerignore` entries, and `CLAUDE.md` describing soldeer deps as git submodules
## What is wrong

`REUSE.toml` and `.soldeerignore` still name `foundry.lock`, `.gitmodules` and `/lib` — none of which exists here — and `CLAUDE.md` still tells the reader the dependencies are git submodules pinned at commit SHAs, updated with `forge update`. All of it is residue from the submodule → soldeer migration.

The `CLAUDE.md` half matters more than usual in this repo: it sits inside a **breaking dependency bump** runbook, so the instruction that is wrong is the one someone follows while changing an audited, deployed stack.

## Everything dangling in this repo

**Submodule paths that no longer exist**

- `REUSE.toml` line 20 — `"foundry.lock",`. There is no `foundry.lock`; it was deleted here already.
- `.soldeerignore` line 12 — `.gitmodules`. There is no `.gitmodules` in the tree and `git ls-files --stage` reports zero gitlinks.
- `.soldeerignore` line 25 — `/foundry.lock`, same absent file.
- `.soldeerignore` line 27 — `/lib`. There is no `lib/` directory. `foundry.toml` sets `libs = ['dependencies']`, so forge resolves everything under `dependencies/` and never creates `lib/`.

`foundry.lock` is Foundry's **git submodule** lockfile — it records the commit each dependency vendored under `lib/` is pinned to, so `forge install` / `forge update` can restore identical submodule revisions. It is only meaningful in a repo that vendors dependencies as git submodules, and this repo does not: `soldeer.lock` is the live lockfile for 13 packages, all under `dependencies/`.

None of it can come back. rainix CI runs a [`no-submodules`](https://github.com/rainlanguage/rainix/blob/main/.github/actions/no-submodules/action.yml) check that fails on a root `.gitmodules` or any committed gitlink.

**Stale documentation asserting the same dead layout**

- `CLAUDE.md` line 171 — *"Git submodules managed via Foundry. Key remappings in `foundry.toml`:"* followed by `rain.vats/`, `rain.factory/`, `rain.deploy/`, `rain.sol.codegen/`, `openzeppelin-contracts-upgradeable/` and `rain.math.float/`.

  None of these are submodules and none of those remapping prefixes exists. `foundry.toml`'s only `remappings` entry is the `@openzeppelin/contracts/` bridge; the real prefixes are the versioned soldeer ones in `remappings.txt` (`rain-vats-0.1.6/`, `rain-factory-0.1.1/`, `rain-deploy-0.1.4/`, `rain-sol-codegen-0.1.0/`, `@openzeppelin-contracts-upgradeable-5.6.1/`, `rain-math-float-0.1.1/`).

- `CLAUDE.md` line 183 — *"Two submodule bumps are **breaking** …"*
- `CLAUDE.md` line 248 — *"Both dependencies are submodules pinned at a specific commit SHA; a `forge update` without the follow-up verification steps above is NOT safe on this stack."*

  These are soldeer dependencies pinned to **semver versions**, not commit SHAs, and `forge update` is not how they move — `forge soldeer update` is, and it rewrites the versioned import prefixes. The safety warning itself is correct and should be kept; the mechanism it names is wrong, so anyone following it literally reaches for a command that does nothing here and skips the one that does.

**A further dangling `REUSE.toml` annotation, which also goes**

- `REUSE.toml` line 22 — `".claude/**/",`. No `.claude/` directory is tracked. (`.gitignore` covers `.claude/scheduled_tasks.lock` only.)

**Absent, but not residue — these `.soldeerignore` entries stay**

- `.soldeerignore` line 29 — `/prod`. No `prod/` directory, and no `.gitignore` entry for one.
- `.soldeerignore` line 35 — `/meta`. No `meta/` directory, and no `.gitignore` entry for one.
- `.soldeerignore` lines 2, 3, 4, 6, 14 — `.cargo`, `.coderabbit.yaml`, `.devcontainer.json`, `.envrc`, `.prettierignore`. This `.soldeerignore` was copied from a Rust/JS-shaped repo; none of these files exists here and none is gitignored, so nothing produces them. There is no Rust in this repo — no `Cargo.toml`, no `crates/`.

Nothing produces any of them *today*, which is not the same as never. Every one of these entries is a live forward-looking filter and none of these lines is removed; they are listed only so they are not mistaken for residue.

**Deliberately not in scope.** `.soldeerignore` also names `.DS_Store`, `.env`, `.fixes`, `.vscode`, `.pre-commit-config.yaml` and the build/publish outputs (`/out`, `/cache`, `/dependencies`, `/deployments`). Those are absent from a clean checkout **by design** — OS junk, local developer files, or artifacts generated at `forge soldeer install` / `forge build` time and therefore present when `soldeer push` runs. They are correct ignores and must stay.

**Nothing else references any of it.** `.github/workflows/`, `flake.nix`, `foundry.toml`, `remappings.txt`, `.gitignore`, `script/`, `docs/`, `CHANGELOG.md`, `README.md` and the rest of the tree were grepped and are clean.

### Which of these actually get removed

`REUSE.toml` and `.soldeerignore` fail in opposite directions, so the same absent path is residue in one and a live rule in the other.

`REUSE.toml` **describes files that exist**. An annotation naming a path that is not there describes nothing, so it is residue and it goes — that is both `foundry.lock` and `.claude/**/` above. `reuse lint` (the `rainix-sol` `legal` job) tolerates a dangling annotation, which is why they survived; what it does not tolerate is the reverse, dropping the annotation of a file that *is* there. Proved live in rain.lib.memkv — removing the annotation for the tracked `soldeer.lock` makes `reuse lint` exit 1.

`.soldeerignore` is a publish **filter**, not a description of the tree. An entry naming an absent path is a forward-looking rule: it is the thing that keeps that path out of the published package if anyone ever adds it. So the test for deleting an entry is **impossible, not absent** — an entry may only go if the path it names can never exist again.

Lines 12, 25 and 27 pass that test, and they pass it structurally rather than by observation: rainix CI's [`no-submodules`](https://github.com/rainlanguage/rainix/blob/main/.github/actions/no-submodules/action.yml) check fails the build on a root `.gitmodules` or any committed gitlink, `/lib` is that same vendored-submodule directory, and `foundry.lock` is only ever written for `lib/` submodule dependencies. While that check stands none of the three can be recreated, so their filter entries can never match anything again. They are genuinely dead, and they are the only `.soldeerignore` lines this issue removes.

Every other entry — `.cargo`, `.coderabbit.yaml`, `.envrc`, `/prod`, `/meta`, whatever else this file lists — is either matching something today or merely **not yet added**. Deleting one of those cleans nothing up; it arms a leak, so that the day someone adds the file it ships inside the published soldeer package. Proved live in rain.lib.memkv — dropping the `/soldeer.lock` line took `forge soldeer push --dry-run` from 14 entries to 15, with `soldeer.lock` now inside the package.

Nothing catches an over-deletion. No CI job reads `.soldeerignore`, so the mistake goes green and only surfaces later, when someone adds the file and it silently publishes. On this repo that would mean shipping the file inside a published package for an audited, deployed stack.

## Context

This is the tail of an unfinished migration, not a convention.

Across the repos that consume `rain.solmem` via soldeer, nine have already dropped `foundry.lock` (raindex, rain.verify, rain.tofu.erc20-decimals, rain.flare, rain.dia, rain.erc4626.words, rainlang, rain.merkle, **S01-Issuer/st0x.deploy**) and eight still carry one (rain.solmem, rainlang.interface, rain.extrospection, rain.math.float, rain.string, rain.lib.memkv, rain.intorastring, rain.datacontract), each with its own issue.

Deleting the file was never the whole job — every one of those nine left dangling `.gitmodules` and/or `lib/` references behind. This issue covers **st0x.deploy** only.

This is documentation and config only. No Solidity source, no deployed bytecode and no audited artifact changes.

## Done when

- [ ] `REUSE.toml` lines 20 and 22 removed
- [ ] `.soldeerignore` lines 12, 25 and 27 removed
- [ ] every other `.soldeerignore` line left exactly as it is, the template-copied ones (2, 3, 4, 6, 14, 29, 35) included
- [ ] `CLAUDE.md` lines 171, 183 and 248 rewritten against the soldeer layout — versioned pins, `forge soldeer update`, and the real `remappings.txt` prefixes — keeping the breaking-bump safety warning intact
- [ ] no reference to `.gitmodules`, `lib/`, `foundry.lock` or `forge update` remains anywhere in the tree outside `dependencies/`
- [ ] CI green (`rainix-sol` test / static / legal)

thedavidmeister:
Closes https://github.com/S01-Issuer/st0x.deploy/issues/311

Documentation, ignore and annotation files only. No Solidity behaviour, no pins,
no deployed bytecode: `remappings.txt`, `soldeer.lock` and `foundry.toml` are
untouched, and the only file under `test/` that changes is a natspec comment.

## Precondition checked first

`git ls-files --stage | grep ^160000` returns nothing — zero gitlinks — so every
`.gitmodules` reference here was already dead. rainix CI's
[`no-submodules`](https://github.com/rainlanguage/rainix/blob/main/.github/actions/no-submodules/action.yml)
check (pulled in by `rainix-sol-static`) keeps it that way.

## `.soldeerignore` — only the three submodule-era lines

Removed: `.gitmodules`, `/foundry.lock`, `/lib`.

These are the only entries that name the dead layout. `foundry.lock` is
Foundry's **git-submodule** lockfile — it records the commit each dependency
vendored under `lib/` is pinned to, and is meaningless in a repo that installs
via soldeer; `soldeer.lock` is the live lockfile for 13 packages.
`foundry.toml` sets `libs = ['dependencies']`, so forge never creates `lib/`.
All three are unreachable states, not absent-for-now ones.

**Everything else stays, including entries whose path does not exist today.**
`.soldeerignore` is a publish *filter*, not a description of the tree — an entry
for an absent file is a forward-looking rule, and deleting it means that file
ships in the soldeer package the moment someone adds it. So `.cargo`,
`.coderabbit.yaml`, `.devcontainer.json`, `.envrc`, `.prettierignore`, `/prod`
and `/meta` are all kept even though they are template-copied and match nothing
right now. A template-copied ignore still ignores the right thing when the file
appears, and on a `.deploy` repo holding audited code, narrowing what the filter
catches changes what ships.

This satisfies the issue's third checkbox — "every other `.soldeerignore` line
left exactly as it is, the template-copied ones (2, 3, 4, 6, 14, 29, 35)
included" — which asks for exactly this retention, for the reason the issue
itself gives: "none of these lines is removed; they are listed only so they are
not mistaken for residue."

No entry needed spelling correction: `.coderabbit.yaml`, `.devcontainer.json`,
`.envrc`, `.cargo` and `.prettierignore` are all the canonical filenames for the
tools they name.

## `REUSE.toml` — the opposite case

REUSE annotations describe files that actually exist, so a dangling path there is
genuine residue with no forward-looking value.

- `foundry.lock` — already deleted from this repo.
- `.claude/**/` — nothing tracked under it (`.gitignore` covers
  `.claude/scheduled_tasks.lock` only, and gitignored files are outside `reuse
  lint`'s view anyway).

`reuse lint` still reports **261/261** files with copyright and license
information after removing both, so neither annotation was load-bearing.

## `CLAUDE.md` — the part that mattered

The dependency section claimed *"Git submodules managed via Foundry. Key
remappings in `foundry.toml`"* and then listed prefixes (`rain.vats/`,
`rain.factory/`, `rain.sol.codegen/`, …) that resolve to nothing. The real
prefixes are the versioned soldeer ones generated into `remappings.txt`
(`rain-vats-0.1.6/`, `rain-factory-0.1.1/`, …). `foundry.toml`'s single
`remappings` entry is unrelated — it bridges the unversioned
`@openzeppelin/contracts/` prefix that a third-party dependency's own source
hard-codes.

The **breaking dependency bump** runbook is where this was worst. It told the
reader both dependencies are *"submodules pinned at a specific commit SHA"* and
that the command to watch out for is `forge update`. Neither is true, and the
consequence is not cosmetic: someone bumping the OZ ERC-7201 storage pins or
Rain Float's rounding behaviour on an audited, deployed stack would reach for a
command that does nothing here and never find the one that does.

Corrected to: exact semver pins in `foundry.toml`'s `[dependencies]`, moved with
`forge soldeer update`; and because the version is part of the import prefix, a
bump also changes every `@openzeppelin-contracts-upgradeable-5.6.1/` and
`rain-math-float-0.1.1/` import across `src/`, `test/` and `script/`. Soldeer
regenerates `remappings.txt` but does not rewrite those imports (confirmed
against `forge soldeer update --help`: its only source-affecting flag is
`-g, --regenerate-remappings`), so they are updated by hand in the same commit.

**The safety warning itself is unchanged in substance** — all three verification
checklists (ERC-20 layout, ERC-1155 layout, Rain Float precision) are kept
verbatim; only the mechanism they name is fixed.

`forge update` and `submodule` still appear in `CLAUDE.md`, deliberately: as the
negative statements that tell the reader those are *not* how this repo works.
That is the correction the issue asked for, not residue.

## One item beyond the issue's enumeration

The issue states the rest of the tree was grepped and is clean. It is not quite:
`test/src/lib/LibERC20Storage.t.sol` carried the same wrong mechanism in its
file-level natspec — *"If a future `forge update` of
`openzeppelin-contracts-upgradeable` reorders the struct…"*. That is exactly the
file the CLAUDE.md ERC-20 runbook sends the reader to at step 2, so leaving it
would have re-planted the instruction being removed, and the issue's own
acceptance criterion is tree-wide. Fixed as a comment-only edit — two words,
zero bytecode impact, test file, no pin touched.

## Out of scope, flagged for a separate decision

`CLAUDE.md` line 273 (unchanged here) says *"Optimizer: 5000 runs"*;
`foundry.toml` sets `optimizer_runs = 2000`, lowered from 5000 to fit the
audit-H01 `_totalSupply` sync. The same CLAUDE.md bullet says the two "must stay
in sync", so it is self-refuting today. Same family of stale docs, but it is not
submodule residue, so it is not in this PR.

## QA

- **Discriminating tests:** n/a for new tests — the diff adds no executable
  behaviour. The change is instead gated by two existing invariants that do
  discriminate, both run against this branch:
  - `reuse lint` (the `rainix-sol` `legal` job) — **261/261** files carry
    copyright + license info after the `REUSE.toml` deletions. This is the check
    that fails if a removed annotation was actually covering a tracked file.
  - `no-submodules` (inside `rainix-sol-static`) — fails on a root `.gitmodules`
    or any committed gitlink, which is the standing guarantee that the three
    removed `.soldeerignore` entries can never become live again.
  - Verified on base: `git ls-files --stage | grep ^160000` → empty on
    `origin/main` as well as on this branch, so the precondition held before the
    edit, not just after.
- **Mutations applied:** n/a — docs/ignore/annotation-only diff with no
  executable lines to mutate. The nearest equivalent was applied and reverted:
  removing the seven non-submodule `.soldeerignore` entries (`.cargo`,
  `.coderabbit.yaml`, `.devcontainer.json`, `.envrc`, `.prettierignore`, `/prod`,
  `/meta`) was tried in 962c569 and **survived every check** — `reuse lint`,
  `forge build` and the pre-commit suite are all blind to it, because a publish
  filter is only exercised at `soldeer push`. That surviving mutant is precisely
  why those lines are restored in 707d215 rather than trusted to CI: nothing in
  this repo's test suite would have caught the package shipping `.envrc` or
  `/prod`.
- **Oracle:** the repo's own live state, read independently of the prose being
  corrected — `foundry.toml` `[dependencies]` (13 exact semver pins) and
  `libs = ['dependencies']`; `remappings.txt` (the 13 real versioned prefixes);
  `soldeer.lock`; `git ls-files --stage` for gitlinks; `forge soldeer update
  --help` for what the update command does and does not rewrite; and
  `rainlanguage/rainix@main` `rainix-sol-static.yaml` for the `no-submodules`
  claim. Every corrected sentence was derived from those, not from the old text.
- **Category check:** issue asks for (a) `REUSE.toml` 20 + 22, (b)
  `.soldeerignore` 12, 25, 27, (c) every other `.soldeerignore` line left exactly
  as it is, template-copied ones included, (d)
  `CLAUDE.md` 171/183/248 rewritten with the safety warning intact, (e) tree-wide
  zero references, (f) CI green. Covered: **a, b, c, d, e, f** — plus the
  `LibERC20Storage.t.sol` natspec that (e) requires but the issue's enumeration
  missed. (c) is satisfied by retention: `.soldeerignore` is +0/-3, so the three
  submodule-era lines are the only ones that moved and every template-copied
  entry stands untouched. Category, not example list: every `.soldeerignore` entry was
  re-checked against "is this path unreachable, or merely absent today?", which
  is the property that separates the three removed lines from the rest.

## Summary by CodeRabbit

* **Documentation**
  * Updated dependency-management guidance to use Soldeer with exact version pins, dependency installation paths, and versioned imports.
  * Clarified package update procedures and CI expectations.
  * Updated regression-test instructions to use the current dependency update command.

* **Chores**
  * Refined repository ignore and licensing annotation configurations.


thedavidmeister:
🤖 ai:vetter
vet-protocol 4
lens source@108fdeba7bc1d728d1c94dbe61917aeed4be33ee + audit skill invoked at pr:312
Reviewed 108fdeba7bc1d728d1c94dbe61917aeed4be33ee: ready — Closes #311 — .soldeerignore drops exactly .gitmodules//foundry.lock//lib (every other filter line kept verbatim), REUSE.toml's two dangling annotations gone, CLAUDE.md dependency+breaking-bump runbook now matches the live soldeer config (13 exact pins, forge soldeer update, versioned import prefixes; all three verification checklists intact and their named tests exist), tree-wide sweep leaves only negative-statement mentions, and the .t.sol edit is natspec-only (=0.8.25 concrete test contract conforms). Stale "Optimizer: 5000 runs" at CLAUDE.md:273 vs optimizer_runs=2000 is untouched-by-diff residue the PR explicitly defers.
cost 118 — docs and publish-filter residue cleanup

PR #310, opened 2026-08-14
Pin tCBRS across Base, Ethereum and HyperEVM (41 -> 42)

alastairong1:
Cerebras Systems (CBRS) launched 2026-08-14. Deployed on Base via `sft-ops create-sft`, then copied to Ethereum and HyperEVM by `20260807-deploy-missing-tokens` dispatched from this branch.

| Chain | receipt | receiptVault | wrapper |
| --- | --- | --- | --- |
| Base | `0xAe8aD234…F14FB` | `0xBeB0c201…c5D1D` | `0xB457cfBF…820ca` |
| Ethereum | `0x8Ea1ba9F…4AEA8` | `0x75E0d127…cb37C` | `0x15925E1c…23962` |
| HyperEVM | `0x8Ea1ba9F…4AEA8` | `0x75E0d127…cb37C` | `0x15925E1c…23962` |

Broadcast runs: [31845108154](https://github.com/S01-Issuer/st0x.deploy/actions/runs/31845108154) (ethereum), [31845492796](https://github.com/S01-Issuer/st0x.deploy/actions/runs/31845492796) (hyperevm). Each selected exactly one token — "Copying 1 Base tokens onto chain id …".

## Verification

Every address read from the live contracts before pinning. On all three chains: `name()` = "Cerebras Systems Inc. ST0x", `symbol()` = `tCBRS`, `authorizer()` = that chain's V4 clone, `owner()` = that chain's token-owner Safe, `receipt()` matching the pinned receipt. Base additionally `isCertificationExpired()` = false (certified to 2027-08-14).

Green at 42 rows: `testConfigMatchesLiveBase`, `testWrappedDerivationHoldsOnBase`, `testConfigAlignsWithBaseTokenTable`, `testEthereumTokenTableMirrorsBaseUnderlyings`, `testProdReceiptVaultsUniformOwnership`, `testProdReceiptVaultsShareUniformAuthoriser`, and the 22 copy-script selection tests.

Sequenced deploy → pin Base → copy → pin targets on one branch, so `main` never carries a red parity window.

## Note

As with every token on the non-Base chains, CBRS is **certified on Base only**. `st0x.deploy` has no certification path, and `certify` is owner-gated with ownership handed to the Safe in the same broadcast — so ordinary transfers on Ethereum and HyperEVM will revert `CertificationExpired` until the Safe certifies. Pre-existing condition affecting all 42, not introduced here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #337, opened 2026-09-03
Pin CBRS, the EU batch and NKE/MCD into the Base tables

alastairong1:
## What

| table | before | after |
| --- | --- | --- |
| `LibProdTokenConfig.productionTokenConfigs()` | 41 | **50** |
| `LibTokenInvariants.productionTokensBase()` | 41 | **50** |
| `productionTokensEthereum()` | 41 | **42** |
| `productionTokensHyperEvm()` | 41 | **42** |

### Base — nine rows

| idx | underlying | deployed | note |
| --- | --- | --- | --- |
| 41 | `CBRS` | 2026-08-14 | shipped alongside FTF; only FTF got pinned |
| 42–47 | `AIR.PA` `BMW.DE` `MC.PA` `SIE.DE` `MBG.DE` `RHM.DE` | 2026-08-28 | EU batch |
| 48–49 | `MCD` `NKE` | 2026-09-03 | today |

### Ethereum + HyperEVM — one row, and this is the important one

**CBRS is already deployed on both target chains.** `20260807-deploy-missing-tokens` copied it on 2026-08-14 ([31845108154](https://github.com/S01-Issuer/st0x.deploy/actions/runs/31845108154) ethereum, [31845492796](https://github.com/S01-Issuer/st0x.deploy/actions/runs/31845492796) hyperevm). The pin PR for it — **#310** — was opened the same day and never merged, so both target tables have sat a row behind the chain ever since.

Pinning CBRS on Base *alone* would have been actively worse than leaving it out. `_selectMissing` decides what to deploy by diffing Base against the **pinned** target table, not against the chain. A Base-only CBRS row makes the copy conclude CBRS is missing on Ethereum and HyperEVM — and **deploy it a second time**. So both target tables get the CBRS row here.

Verified live on both chains before pinning:

```
vault   0x75E0d127794b9C26eE35c55fbaBcc41c53Ccb37C
          symbol()  tCBRS
          name()    Cerebras Systems Inc. ST0x
          owner()   0x3840aeDa…0329   the token-owner Safe
receipt 0x8Ea1ba9Fc0CF7338B41DdDa5B778a9118274AEA8   from receipt() on the vault
wrapper 0x15925E1c19c0F0d392F6FCb40FdE9144Dd823962   symbol() wtCBRS
```

**#310 is now redundant and should be closed rather than merged** — it pins CBRS at index 41 of a Base table this branch already takes to 50.

## Where the Base addresses came from

Read off Base, not copied from the ledger. `ops/launches.json` records only `sft` and `wrapper`; the **receipt** leg is not in it, so each came from `receipt()` on the vault. Every vault also checked for `owner() == 0xe70d821f…d611` and `authorizer() == 0x315b16fa…f0cD` — all nine clean.

The two orphaned `tMBG.DE` deploys from 2026-08-28 (runs `33143416198`, `33143755146`) are `deprecated` in the ledger and are **not** referenced. Row 46 is `0x07d7ca93…73A0`, the third deploy.

## Two things that are new to these tables

**Dotted tickers.** The underlying string keeps the dot — `"AIR.PA"` — because that is what Base reports from `symbol()` minus the `t`. Only the Solidity constant names substitute an underscore (`AIR_PA_RECEIPT`); a dot is not a valid identifier character.

**A non-ASCII name.** `MC.PA` is `LVMH Moët Hennessy Louis Vuitton SE ST0x`. solc rejects the bare literal (`Error 8936`), so that row uses `unicode"..."`. Dropping the `ë` would just fail `testConfigMatchesLiveBase` instead — Base stores the accent.

## Verification

```
testConfigAlignsWithBaseTokenTable       PASS   index pairing, config <-> Base
testConfigMatchesLiveBase                PASS   name + symbol, all 50, live Base
testWrappedDerivationHoldsOnBase         PASS
testAssertProductionStateBasePassesLive  PASS   owner / authoriser / wiring
```
plus 627 non-fork tests.

## Parity is red, and this PR should NOT merge that way

```
testCrossChainParity  FAIL: token table lengths diverge: 50 != 42
```

That is the correct number. Base 50, targets 42 — the gap is exactly the eight tokens the copy has yet to deploy (EU batch + NKE/MCD). CBRS is no longer part of the gap, which is the point of the second commit.

**But red is a transient state for this branch, not for `main`.** #310 established the pattern and said so explicitly: *"Sequenced deploy → pin Base → copy → pin targets on one branch, so `main` never carries a red parity window."* `manual-broadcast` is a `workflow_dispatch`, so it can be dispatched against this branch.

## Sequence — all on this branch, before merge

1. Dispatch `manual-broadcast` against **`2026-09-03-pin-cbrs-eu-nke-mcd`**, `script = 20260807-deploy-missing-tokens`, `network = ethereum`
2. Same again, `network = hyperevm`
3. `_selectMissing` should report **8 tokens per chain** — if it says 9, CBRS was not picked up from the second commit and the run must be stopped before it redeploys it
4. Pin each run's logged `(underlying, receipt, receiptVault, wrapped)` tuples into `productionTokensEthereum()` / `productionTokensHyperEvm()`, taking both to 50
5. Parity goes green → merge

Merging at step 0 would put a red parity window on `main` and, worse, leave the two target tables 8 rows behind the chain — the same failure mode as CBRS, at eight times the size.

### DIFF
--- a/src/lib/LibProdTokenConfig.sol
+++ b/src/lib/LibProdTokenConfig.sol
@@ -42,10 +42,10 @@ struct TokenConfig {
 /// means carrying that space forward; the parity pin would flag it as a
 /// divergence otherwise.
 library LibProdTokenConfig {
-    /// @notice The 41 production token deploy configs, Base table order.
+    /// @notice The 50 production token deploy configs, Base table order.
     /// @return configs The name/symbol table.
     function productionTokenConfigs() internal pure returns (TokenConfig[] memory configs) {
-        configs = new TokenConfig[](41);
+        configs = new TokenConfig[](50);
         configs[0] = TokenConfig("MSTR", "MicroStrategy Incorporated ST0x", "tMSTR");
         configs[1] = TokenConfig("TSLA", "Tesla Inc ST0x", "tTSLA");
         configs[2] = TokenConfig("COIN", "Coinbase Global Inc ST0x", "tCOIN");
@@ -88,5 +88,16 @@ library LibProdTokenConfig {
         configs[38] = TokenConfig("BABA", "Alibaba Group Holding Limited ADR ST0x", "tBABA");
         configs[39] = TokenConfig("TQQQ", "ProShares UltraPro QQQ ST0x", "tTQQQ");
         configs[40] = TokenConfig("FTF", "Franklin Limited Duration Income Trust ST0x", "tFTF");
+        configs[41] = TokenConfig("CBRS", "Cerebras Systems Inc. ST0x", "tCBRS");
+        configs[42] = TokenConfig("AIR.PA", "Airbus SE ST0x", "tAIR.PA");
+        configs[43] = TokenConfig("BMW.DE", "Bayerische Motoren Werke Aktiengesellschaft ST0x", "tBMW.DE");
+        // solc rejects a bare non-ASCII string literal; `unicode"..."` is the
+        // same bytes, and Base carries the ë — dropping it would break parity.
+        configs[44] = TokenConfig("MC.PA", unicode"LVMH Moët Hennessy Louis Vuitton SE ST0x", "tMC.PA");
+        configs[45] = TokenConfig("SIE.DE", "Siemens Aktiengesellschaft ST0x", "tSIE.DE");
+        configs[46] = TokenConfig("MBG.DE", "Mercedes-Benz Group AG ST0x", "tMBG.DE");
+        configs[47] = TokenConfig("RHM.DE", "Rheinmetall AG ST0x", "tRHM.DE");
+        configs[48] = TokenConfig("MCD", "McDonald's Corporation ST0x", "tMCD");
+        configs[49] = TokenConfig("NKE", "NIKE, Inc. ST0x", "tNKE");
     }
 }
--- a/src/lib/LibTokenInvariants.sol
+++ b/src/lib/LibTokenInvariants.sol
@@ -534,7 +534,7 @@ library LibTokenInvariants {
     /// @notice Returns the production token instance triples on Ethereum
     /// mainnet — Base's underlyings in Base row order, so the tables pair by
     /// index as well as by key.
-    /// @return tokens The 41 production token instances on Ethereum.
+    /// @return tokens The 42 production token instances on Ethereum.
     function productionTokensEthereum() internal pure returns (TokenInstance[] memory tokens) {
         // Deployed on Ethereum mainnet 2026-07-22 by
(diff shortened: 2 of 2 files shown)


alastairong1:
### Reading the red `rainix-sol / test`

**Exactly one failure belongs to this PR:**

```
[FAIL: Ethereum token table length diverges from Base: 42 != 50]
  testEthereumTokenTableMirrorsBaseUnderlyings()
```

That is the intended intermediate state — Base 50, targets 42, gap = the eight tokens the copy has yet to deploy. It closes when the dispatch runs and the tuples are pinned onto this branch.

**Everything else is pre-existing on `main`.** The `BeaconOwnerMismatch` / `UnexpectedBeaconOwner` / `MigrationStateDrift` failures all report the same thing — beacon `0xb9DCd744…4613` is owned on chain by `0x8E4bdeec7CEB9570D440676345dA1dCe10329f5b`, while the tests expect the token-owner Safes (`0xe70d821f…d611` on Base, `0x3840aeDa…0329` on the targets):

```
testFleetRolloutBase, testFleetRolloutEthereum, testFleetRolloutHyperEvm,
testProdDeployBaseV4, testBeaconUpgradeRunsThroughTimelock,
testGovernanceCanMigrateAwayToSuccessor, testRunAuthorsFullMigration{,OnEthereum,OnHyperevm},
testRunRefusesWhenNothingToMigrate, testSetAuthorizerRunsThroughTimelock,
testVerifyAcceptsAuthoredArtifactAndRejectsTamper,
test{Base,Ethereum,Hyperevm}GovernanceInMigrationWindow
```

`main` has been red on this since [run 33765524907](https://github.com/S01-Issuer/st0x.deploy/actions/runs/33765524907) (the #336 merge, ~14:12 today) — it is unrelated to this branch and worth someone picking up separately. It looks like an on-chain governance handoff the repo hasn't caught up with.

So the review question here is only whether the 50 Base rows and the two CBRS target rows are right, not the CI colour.

alastairong1:
Superseded by #339.

#339 branches off this branch (`2026-09-03-pin-cbrs-eu-nke-mcd`) and carries everything here forward unchanged, then extends the Base table from 50 to 56 with the six launched tokens deployed since — GRND (2026-09-04) and the 2026-09-06 batch DNUT / PLBY / TR / WEN / FGI — plus commented-out records for the three deployed-but-never-launched tokens (tMETA, tPLTR, tGM).

Please review and merge #339 instead; this PR can be closed once #339 lands. Nothing has been merged from either branch.

